PR #41, opened 2019-02-02
Steps, bicycle ramp, and bigger cycleway.

Florimondable:
Steps have it own style, with bicycle ramp.
Cycle style doesn't depend on pedestrian anymore.
Cycle is bigger around Z11-15.
cyclosm_steps_ramp_cycle_z18
In this image you have steps (upper) and other steps with a bicycle ramp (lower).

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -243,9 +243,14 @@ Layer:
           WHEN highway IN ('motorway_link', 'trunk_link', 'primary_link', 'secondary_link', 'tertiary', 'tertiary_link', 'residential', 'unclassified', 'road', 'living_street') THEN 'minorroad'
           WHEN highway IN ('service', 'track') THEN 'service'
           WHEN highway='cycleway' THEN 'bike'
-          WHEN highway IN ('path', 'footway', 'pedestrian', 'steps', 'bridleway') THEN 'pedestrian'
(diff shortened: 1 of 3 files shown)


project.mml
```diff
@@ -287,12 +292,16 @@ Layer:
           WHEN highway IN ('motorway_link', 'trunk_link', 'primary_link', 'secondary_link', 'tertiary', 'tertiary_link', 'residential', 'unclassified', 'road', 'living_street') THEN 'minorroad'
           WHEN highway IN ('service', 'track') THEN 'service'
           WHEN highway='cycleway' THEN 'cycle'
-          WHEN highway IN ('path', 'footway', 'pedestrian', 'steps', 'bridleway') THEN CASE WHEN bicycle IN ('yes', 'designated') THEN 'cycle' ELSE 'pedestrian' END
+          WHEN highway IN ('path', 'footway', 'pedestrian', 'bridleway') THEN CASE WHEN bicycle IN ('yes', 'designated') THEN 'cycle' ELSE 'pedestrian' END
```
Phyks: Shouldn't this test with bicycle also be done in bridge and tunnels?

Florimondable: We shouldn't mixed pedestrian, bridleway and footway with cycleway rendering, they could become blue if they are bicycle friendly though.

Issue #12, opened 2019-01-26
Render surface tags?


Phyks:
Should `surface` tags be rendered (`paved`/`unpaved`)? What could be a good way to render this?

Phyks:
https://tilemill-project.github.io/tilemill/docs/guides/styling-lines/ line-pattern-file

could be an idea to display a "paved road" effect.

Florimondable:
My thought of how to do it in this discussion :
«Florimond Berthoux [13 h 58]
avec https://tilemill-project.github.io/tilemill/docs/guides/styling-lines/ line-pattern-file il y a moyen d'afficher un effet route pavée
ou alors c'est avec du dash line
l'intérêt du dash line c'est qu'on pourrait rendre avec la même symbolique des différents traitillé les différent revêtements d'une autoroute à un footway
sinon il faudra faire deux symboliques l'une pour les routes avec un effet pavé/caillou, un autre pour les path

Florimond Berthoux [14 h 34]
en réutilisant la symbolique des grade de tracktype sur osm.org https://wiki.openstreetmap.org/wiki/Key:tracktype
en tant que ligne axiale

Lucas Verney [14 h 36]
dash line, ça risque de faire trop de traitillés avec les pistes cyclables, non ?

Florimond Berthoux [14 h 36]
possible
un line pattern file, ça risque aussi de faire très chargé et brouilli
je pense qu'il ne faudrait l'afficher qu'à partir de Z14 (modifié) 
et si c'est du grade1 (revêtement lisse) ne rien affiché

Florimond Berthoux [14 h 45]
il y aurait trois niveaux, suivant le type de vélo utilisable facilement : vélo de route (pneu<=28) - surface lisse, vélo gravel/ville/rando (pneu <=35) surface pavée/caillou fin, vtt (pneu >35) gros caillou/terre à détuire des tags gratype/smoothnees/surface

Lucas Verney [14 h 47]
:+1:

Florimond Berthoux [14 h 47]
ce qui fait que deux dash possible : "- - - - -" ou "- . - . - . -" facilement reconnaissable»

sylvainavelo:
The way road surface is rendered is not clear because it uses two different visual variables at the same time : the width of the outline (seems wider than other roads) and the dashed fill.  Only one should be used.

This is not a usual way of rendering road surface on a map. Usually the dash is applied on the outline, not on the fill.

[...]

PR #75, opened 2019-02-25
Basic surface render

Phyks:
Here is a quick test to render `surface=paved` using `line-pattern`.

capture d ecran 2019-02-25 a 15 54 29

The pattern is too strong and not very nice, but I guess we could do something nice out of it. This is just a 5 minutes test.

I was aiming for something like this (quickly drawn in Gimp):

capture d ecran 2019-02-25 a 15 33 36

See #12


Phyks:
Closing in favor of #77, way more advanced and better render so far.

PR #76, opened 2019-02-26
Render pedestrian as a street

Florimondable:
cyclosm_pedestrian_surfaces_1
(don't look at the surface rendering by axial dash line).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,24 @@ designed with cycling in mind.
 
 **Note:** This is currently a Work In Progress.
 
+## Features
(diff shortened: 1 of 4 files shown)


Phyks:
> Add features list to README.md.

:+1: for this. There is also a legend on https://tiles.phyks.me/ which is a bit outdated. Maybe we should put this in a legend file in this repo with the full legend, similarly to the legend on osm.org.

Florimondable:
It is a good idea.

PR #77, opened 2019-02-26
Surface and Pedestrian

### LINKED ISSUE #12: Render surface tags?

Florimondable:
cyclosm_pedestrian_surfaces_1

EDIT: Fixes #12.


Phyks:
:heavy_plus_sign: for the surface classification.

However, I fear the axial line might make too much dashes with cycleways on both side and so on, especially on small roads. :/

Florimondable:
I'm wonder if I should do the dash over the whole road  with a lighter (or darker if white) color

roads.mss
```diff
@@ -2476,7 +2476,7 @@ come in as well.
     [type='service']      { line-width: @rdz13_service; }
     [type='track']      { line-width: @rdz13_track; }
     [type='pedestrian']   { line-width: @rdz13_pedestrian; }
-    [type='bridleway']   { line-width: @rdz13_bridleway; line-dasharray: 1,1;}
+  [type='bridleway']   { line-width: @rdz13_bridleway; /*line-dasharray: 1,1;*/}
```
Phyks: Missing indent here

roads.mss
```diff
@@ -2667,6 +2667,428 @@ come in as well.
...
+    [type='trunk_link'] {
+      line-color: lighten(@trunk-fill,@surfaceLighterDarker1);
+    }
+    [type='primary'],
+    [type='primary_link'] {
+      line-color: lighten(@primary-fill,@surfaceLighterDarker1);
+    }
+    [type='secondary'],
+    [type='secondary_link'] {
+      line-color: lighten(@secondary-fill,@surfaceLighterDarker1);
```
Phyks: Not sure this is visible enough on secondary (cf picture on "Boulevard d'Alsace Lorraine" you sent me). No other concerns with the colors so far.

roads.mss
```diff
@@ -2667,6 +2667,428 @@ come in as well.
...
+    [type='tertiary_link'],
+    [type='secondary_link'],
+    [type='primary_link'],
+    [type='trunk_link'],
+    [type='motorway_link'],
+    [type='primary'],
+    [type='secondary'],
+    [type='motorway'],
+    [type='trunk'] {
+      line-cap: round;
```
Phyks: Not sure we need the line-cap / line-join here (did not check in details). If this is useless, would probably ease a bit the filtering and render.

roads.mss
```diff
@@ -2667,6 +2667,428 @@ come in as well.
...
+      [type='track']      { line-width: @rdz16_track; }
+      [type='pedestrian']   { line-width: @rdz16_pedestrian; }
+      [type='bridleway']   { line-width: @rdz16_bridleway; }
+      [type='footway']   { line-width: @rdz16_footway; }
+      [type='path'],
+      [type='footway'][bicycle='yes'],
+      [type='footway'][bicycle='designated'],
+      [type='bridleway'][bicycle='yes'],
+      [type='bridleway'][bicycle='designated'] {
+        line-width: @rdz16_path;
```
Phyks: All the line widths here could probably be put in a single place with all the line widths below. This would avoid (a bit of) code duplication.

roads.mss
Florimondable: yep

roads.mss
Florimondable: But they must not be rendered if the surface quality good...

Florimondable:
New rendering with the whole with of the road to be dashed.
Did two improvements, better dash distinction between both quality and the color them self are different (10% lighter for middle quality and 20% lighter for bad quality road).

Didn't add every kind of road in the style, cost two much to generate the .xml.
This need to be optimize.
cyclosm_surfaces_8_push

Florimondable:
Thanks for your sql optimization it works well (albeit a bug in it :p).
I think we have now something that can be merged.
cyclosm_surfacequality_paris_1
cyclosm_surfacequality_paris_2

PR #113, opened 2019-03-25
Change cycleway= rendering

Florimondable:
- 2 less road outline render
- cycleway= use same width than highway=cycleway

cyclosm_render_cycleway

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1885,7 +1885,7 @@ come in as well.
       [type='secondary_link']    { line-width: @rdz18_secondary_link; }
       [type='primary_link']    { line-width: @rdz18_primary_link; }
       [type='service']      { line-width: @rdz18_service; }
-      [type='track']      { line-width: @rdz12_track; }  //Special effect.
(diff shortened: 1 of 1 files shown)


roads.mss
```diff
@@ -2377,8 +1885,8 @@ come in as well.
       [type='secondary_link']    { line-width: @rdz18_secondary_link; }
       [type='primary_link']    { line-width: @rdz18_primary_link; }
       [type='service']      { line-width: @rdz18_service; }
-      [type='track']      { line-width: @rdz17_track; }  //Special effect.
-      [type='pedestrian']   { line-width: @rdz18_pedestrian; }
+      [type='track']      { line-width: @rdz12_track; }  //Special effect.
```
Phyks: Is the rdz12 here expected or a typo? (Same below)

Florimondable: must be an error, at one moment I replaced every 17 by 12 😱 , it uses 17 instead of 18 to have a casing effect to let the track visible with white surface dashing.

PR #116, opened 2019-04-03
Improve compilation time

Phyks:
After merging #113, the compilation time became very long (about 10s for each road layer).

This should improve this and simplify a bit the data model. I'm now only using one cycleway variable for each side:
* `cycleway_{left,right}` contains the left (right) cycleway value (or the `cycleway` tag if no overloading is possible with side cycleway tag).
* `has_cycleway_{left,right}` is used as a boolean value to check whether the left (right) outline should be overloaded.

I gave a try and it should be ok. @Florimondable, might I ask you to have a look and double check I didn't mess up this change? Thanks!

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -355,13 +355,22 @@ Layer:
           END AS maxspeed_kmh,
           bicycle, tags->'motor_vehicle' AS motor_vehicle,
           oneway,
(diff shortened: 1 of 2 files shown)


Florimondable:
Looks good.

Florimondable:
May be we could merge opposite_track and track ? (same for busway and lane)

Phyks:
Depends whether we want directional arrows in the future or not :)

Issue #114, opened 2019-04-01
Regional route too transparent


Phyks:
See "NS" here https://cyclosm.github.io/cyclosm-cartocss-style/example.html#17/48.83473/2.33783

PR #117, opened 2019-04-03
Better render of regional routes

Phyks:
Following https://github.com/cyclosm/cyclosm-cartocss-style/issues/114, I think https://github.com/cyclosm/cyclosm-cartocss-style/pull/112/files was too large actually and should only apply to the local routes as this is the dark blue which is too strong and opaque. Purplish / Pinkish colors should be fine at higher zooms.

I applied the opacity change only to local routes then. What do you think about it?

(First commit is just the one from #116)

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -355,13 +355,22 @@ Layer:
           END AS maxspeed_kmh,
           bicycle, tags->'motor_vehicle' AS motor_vehicle,
           oneway,
-          tags->'cycleway' AS cycleway, tags->'cycleway:left' AS cycleway_left, tags->'cycleway:right' AS cycleway_right,
+          CASE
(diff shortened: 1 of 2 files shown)


Florimondable:
Do you have a screenshot ? :p

Phyks:
https://cyclosm.github.io/cyclosm-cartocss-style/example.html#18/48.83941/2.33912

Capture d’écran 2019-04-11 à 10 47 36


https://cyclosm.github.io/cyclosm-cartocss-style/example.html#17/48.87435/2.36623

Capture d’écran 2019-04-11 à 10 48 31

Canal St-Martin is a mess, but regional routes are not visible otherwise. My opinion on this is that we should probably handle route superposition somehow, or draw cycling infra on top of routes, in a separate layer. We'll never manage to fix this issue completely while keeping everything readable just by playing with transparency :/

PR #125, opened 2019-05-08
Render cyclestreet=yes as a 20km/h street.

Phyks:
What do you think about rendering them as 20km/h streets?

In Belgium for instance, they are limited to 30km/h, but given that cars should behave as guests and bicycles should be the main users, I guess rendering them as 20km/h streets (which is an average speed for a typical bike) is fine.

I don't know this infrastructure very well, but it seems there is not much added benefits from a cycle street rather than a regular (well done) 20km/h limitation, isn't it?

The Bicycle tag map renders them as pink streets, but this cannot apply to us as pink is already in use for the bicycle routes.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -347,6 +347,7 @@ Layer:
       (
         SELECT way, COALESCE(highway, CASE WHEN railway IN ('light_rail', 'subway', 'narrow_gauge', 'rail', 'tram') THEN 'railway' ELSE 'other' END) AS type, 0 AS bridge, access, layer, 1 as tunnel,
           CASE
+            WHEN tags->'cyclestreet'='yes' THEN 20 -- render cyclestreets as maxspeed 20km/h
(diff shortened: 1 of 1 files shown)


Florimondable:
Why not, though I don't like to make the legend more complexe.
20km/h means that pedestrian are more welcome on the road which means that the road might not be as suitable for bike as a cyclestreet.

Phyks:
I'm a bit mixed here between adding new colors and features (and making the legend more difficult to read) and mixing together different features in the same render.

An alternative way could be to have a color in between `#bbffff` (speed 30) and `#21ffcc` (speed 20). It might not be well distinguishable, but it would not be a big issue in my opinion.

Florimondable:
I don't like to mix speed and cyclestreet, because speed is already a bad way to think that a road is cycle friendly or not. We should focus more on trafic, not speed, but that an other subject.

In this case a cyclestreet is a street for bicycle so must be render with outline like other streets, but in blue almost like cycleway (but little lighter blue in orde to see its not purely bicycle trafic)

Issue #180, opened 2019-09-26
Spurious waterways rendered


Phyks:
We should not render "La Bièvre" here: https://www.cyclosm.org/#map=19/48.84484/2.36516/cyclosm. Not sure where it is coming from.

![2019-09-26-115228](https://user-images.githubusercontent.com/3856586/65678723-225e7080-e054-11e9-81e8-14e4da12c690.png)

Florimondable:
Can you also check underground water tank https://www.cyclosm.org/#map=19/48.88221/2.27954/cyclosm

Phyks:
Yes, this one probably lacks a tag, something like tunnel=yes or so (maybe location=underground?) at least.

PR #181, opened 2019-09-26
No longer render tunnel waterways.

Phyks:
Fix https://github.com/cyclosm/cyclosm-cartocss-style/issues/180.


Phyks:
Actually we had issues with the Canal Saint-Martin upon exiting Bastille https://www.cyclosm.org/#map=16/48.8548/2.3684/cyclosm.

These are now fixed (separate commit)

Florimondable:
Yep, two watreway : riverbank and canal.

Issue #143, opened 2019-06-25
Path with bicycle=designated


Florimondable:
A Path with bicyle=designated is renderd like a normal path. Some cycle track are modelised with tags 
highway=path
bicycle=designated
segregated=yes
This means there is a cycle track so we should render them like normal highway=cycleway

If segregated is not present or is =no we should render them more like a cyleway (less green more blue, as wide as cycleway (with the management of oneway)).

Florimondable:
This ticket should be more on how to render green way ( https://wiki.openstreetmap.org/wiki/FR:Bicycle#Voies_vertes )
May be we should render path depending on there width also.

Dardy87:
confirm, the path rules seems too generic, below an example:
https://www.openstreetmap.org/way/132088570
It's a cycleway shared with pedestrian, it should be rendered near to blue.

Instead below a path for bicycle and pedestrian, but the cycleway is segregated
https://www.openstreetmap.org/way/557624769

Both are rendered in the same way (dark green) but they should result as cycleways, not paths

Phyks:
@Florimondable I think this one can be closed after your latest PR (#184). Isn't it?

Florimondable:
I hope so :D

PR #184, opened 2019-09-26
Render path with bicycle=designated almost as a cycleway and segregated as a cycleway

Florimondable:
link to #143 (but not completed yet, segregated must be done also)

Can't be able to fix the oneway arrow not rendered issue on path with _oneway:bicycle=yes_

Capture d’écran de 2019-09-26 22-36-27

### DIFF
--- a/labels.mss
+++ b/labels.mss
@@ -533,12 +533,13 @@
      marker-max-error: 0.5;
      marker-spacing: 100;
      marker-fill: #777777;
-     [oneway != 'yes'][oneway_bicycle = 'yes'] {
-       marker-fill: #00f;
-     }
      [highway='cycleway'],
      [highway='path'] {
        marker-fill: #ddf;
+     }
+     /* Weird case where the road is oneway for only the bicycles, blue arrow. */
+     [oneway != 'yes'][oneway_bicycle = 'yes'][highway!='path'][highway!='cycleway'] {
+       marker-fill: #00f;
      }
      marker-file: url(symbols/oneway.svg);
      [oneway='-1'] {
--- a/labels.mss
+++ b/labels.mss
@@ -536,7 +536,8 @@
      [oneway != 'yes'][oneway_bicycle = 'yes'] {
        marker-fill: #00f;
      }
-     [highway='cycleway'] {
+     [highway='cycleway'],
+     [highway='path'] {
        marker-fill: #ddf;
      }
      marker-file: url(symbols/oneway.svg);
@@ -549,7 +550,7 @@
      }
   }
   [oneway = 'yes'][oneway_bicycle = 'no'],
-  [oneway='-1'][oneway_bicycle = 'no'] {
+  [oneway = '-1'][oneway_bicycle = 'no'] {
      bike/marker-placement:line;
      bike/marker-max-error: 0.5;
      bike/marker-spacing: 100;
--- a/project.mml
+++ b/project.mml
@@ -523,6 +523,10 @@ Layer:
             ELSE NULL
           END AS designated,
           CASE
+            WHEN tags->'segregated' IN ('yes') THEN 'yes'
+            ELSE 'no'
+          END AS segregated,
+          CASE
             WHEN tags->'oneway:bicycle' IN ('yes', '-1') THEN 'yes'
             ELSE 'no'
           END AS oneway_bicycle,
@@ -639,6 +643,10 @@ Layer:
             WHEN bicycle IN ('designated') THEN 'yes'
             ELSE NULL
           END AS designated,
+          CASE
+            WHEN tags->'segregated' IN ('yes') THEN 'yes'
+            ELSE 'no'
+          END AS segregated,
           CASE
             WHEN tags->'oneway:bicycle' IN ('yes', '-1') THEN 'yes'
             ELSE 'no'
@@ -798,6 +806,10 @@ Layer:
             WHEN bicycle IN ('designated') THEN 'yes'
             ELSE NULL
           END AS designated,
+          CASE
+            WHEN tags->'segregated' IN ('yes') THEN 'yes'
+            ELSE 'no'
+          END AS segregated,
           CASE
             WHEN tags->'oneway:bicycle' IN ('yes', '-1') THEN 'yes'
             ELSE 'no'
--- a/project.mml
+++ b/project.mml
@@ -518,6 +518,10 @@ Layer:
             WHEN access='no' AND (bicycle IS NULL OR bicycle='no') THEN 'no'
             ELSE NULL
           END AS can_bicycle,
+          CASE
+            WHEN bicycle IN ('designated') THEN 'yes'
+            ELSE NULL
+          END AS designated,
           tags->'ramp:stroller' AS ramp_stroller,
           tags->'ramp:bicycle' AS ramp_cycle,
           tags->'ramp:wheelchair' AS ramp_wheelchair,
@@ -627,6 +631,10 @@ Layer:
             WHEN access='no' AND (bicycle IS NULL OR bicycle='no') THEN 'no'
             ELSE NULL
           END AS can_bicycle,
+          CASE
+            WHEN bicycle IN ('designated') THEN 'yes'
+            ELSE NULL
+          END AS designated,
           tags->'ramp:stroller' AS ramp_stroller,
           tags->'ramp:bicycle' AS ramp_cycle,
           tags->'ramp:wheelchair' AS ramp_wheelchair,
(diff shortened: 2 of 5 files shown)


Phyks:
> Can't be able to fix the oneway arrow not rendered issue on path with oneway:bicycle=yes

Here is the fix I think https://github.com/cyclosm/cyclosm-cartocss-style/commit/5f29c76fde0c540c3598246b7c32cad6b4c48a4e.

Phyks:
By the way, could you post before/after pictures (or the URL used, so I can go and check the before on cyclosm.org easily)? Thanks!

Florimondable:
Before (bicycle=designated only) :
Capture du 2019-09-28 23-11-48

After 
Capture du 2019-09-28 23-11-08

higway=path bicycle=designated segregated=yes : draw a cycleway with a right outline as a path
Capture du 2019-09-28 23-10-07


higway=footway bicycle=designated segregated=yes : draw a cycleway with a right outline as a footway
Capture du 2019-09-28 23-08-11

Dardy87:
Are you sure on the last example? in my opinion higway=footway bicycle=designated segregated=yes
Is a wrong tagging (even if I found it in some cases). Mixed cycleways and footway  should be tagged as paths. Is it specified in some wiki?

Florimondable:
@Dardy87 I agree with you it is bad tagging.
But I was wondering if it's a scheme use a lot or not.
After few overpass request around the word i seems it's not wildly used, so we can remove it.

Florimondable:
We should remove also special rendering with highway=footway/bridelway and bicycle=designated, it's bad tagging also and we should not promote bad tagging.

Phyks:
`way[highway=footway][bicycle=designated]` is a bad tagging (up to 7% of the `highway=footway` though, https://taginfo.openstreetmap.org/tags/highway=footway#combinations) and is checked by the JOSM validator and Osmose (https://github.com/osm-fr/osmose-backend/blob/ff294d7b5f6289b968d7165e6246f870e46c846a/plugins/Josm_combinations.py#L1461-L1462).

`way[highway=bridleway][bicycle=designated] ` is probably as well although it's much more common (up to 29% of bridleways, https://taginfo.openstreetmap.org/tags/highway=bridleway#combinations) and not checked by any validation tools. I agree this is bad tagging though and I'll try to submit a patch for extra checking in JOSM etc.

labels.mss
```diff
@@ -533,12 +533,14 @@
...
-     [oneway != 'yes'][oneway_bicycle = 'yes'] {
-       marker-fill: #00f;
-     }
-     [highway='cycleway'] {
+     [highway='cycleway'],
+     [highway='path'] {
        marker-fill: #ddf;
      }
+     /* Weird case where the road is oneway for only the bicycles, blue arrow. */
+     [oneway != 'yes'][oneway_bicycle = 'yes'][highway!='path'][highway!='cycleway'] {
```
Phyks: This can simply be `[oneway != 'yes'][oneway_bicycle = 'yes']` if it is put before the previous `[highway='cycleway'], [highway='path']` rule above. Overloading should do the trick.

Why is it a "weird case"? This is just any countraflow bike lane, isn't it?

labels.mss
```diff
@@ -533,12 +533,14 @@
      marker-max-error: 0.5;
      marker-spacing: 100;
      marker-fill: #777777;
-     [oneway != 'yes'][oneway_bicycle = 'yes'] {
-       marker-fill: #00f;
-     }
-     [highway='cycleway'] {
+     [highway='cycleway'],
+     [highway='path'] {
```
Phyks: It might be better to have an `is_primary_bike` value exported in the SQL output and filter on this one. This one, we could also account for some `bicycle=designated` cases or roads without motor traffic.

Might also be better in terms of Mapnik XML file size / etc.

project.mml
```diff
@@ -518,6 +518,18 @@ Layer:
...
+          CASE
+            WHEN bicycle IN ('designated') THEN 'yes'
+            ELSE NULL
+          END AS designated,
+          CASE
+            WHEN tags->'segregated' IN ('yes') THEN 'yes'
+            ELSE 'no'
+          END AS segregated,
+          CASE
+            WHEN tags->'oneway:bicycle' IN ('yes', '-1') THEN 'yes'
```
Phyks: `THEN tags->'oneway:bicycle'` to keep the direction, or we will never need it?

project.mml
```diff
@@ -627,6 +639,18 @@ Layer:
...
```
Phyks: `THEN tags->'oneway:bicycle'` to keep the direction, or we will never need it?

project.mml
```diff
@@ -778,6 +802,18 @@ Layer:
...
```
Phyks: `THEN tags->'oneway:bicycle'` to keep the direction, or we will never need it?

project.mml
```diff
@@ -1987,8 +2023,7 @@ Layer:
             ELSE 'no'
           END AS oneway,
           CASE
-            WHEN tags->'oneway:bicycle'='no' THEN 'no'
-            WHEN tags->'oneway:bicycle'='yes' THEN 'yes'
+            WHEN tags->'oneway:bicycle' IN ('yes', '-1') THEN 'yes'
```
Phyks: `THEN tags->'oneway:bicycle'` to keep the direction, or we will never need it?

roads.mss
```diff
@@ -1710,9 +1754,13 @@ come in as well.
...
           line-width: @rdz17_path;
         }
-        [type='cycleway'] {
+        [type='cycleway'],
+        [type='path'][designated='yes'],
+        [type='footway'][designated='yes'],
+        [type='bridleway'][designated='yes'] {
           line-width: @rdz17_cycle;
-          [oneway='no'] { line-width: @rdz17_cycle*1.5; }
+          [oneway='no'][oneway_bicycle='no']
```
Phyks: We should probably try to avoid this double check here. I should review the exact tags we need, but probably we could fill in the `oneway_bicycle` value with the `oneway` value by default, therefore having:
* `oneway` which would be the general traffic oneway
* `oneway_bicycle` which would be the real oneway status for bikes

roads.mss
```diff
@@ -1245,14 +1257,14 @@ come in as well.
...
     // bridleway, footway, steps are not shown at Z12
-    [type='path'],
-    [type='footway'][can_bicycle='yes'],
-    [type='bridleway'][can_bicycle='yes'] {
+    [type='path'] {
       line-width: @rdz12_path;
     }
-    [type='cycleway'] {
+    [type='cycleway'],
+    [type='path'][designated='yes'] {
```
Phyks: We have designated, segregated and other tags exported. Not sure but we could probably factor them together. :/

labels.mss
Florimondable: No, it is a two way road, but one way for bicycle, weird, may be a bug, may be we should not render it.

labels.mss
Florimondable: _is_primary_bike_ for what tags combination ?

project.mml
Florimondable: I don't understand the question.

Florimondable:
> `way[highway=footway][bicycle=designated]` is a bad tagging (up to 7% of the `highway=footway` though, https://taginfo.openstreetmap.org/tags/highway=footway#combinations) 

I read 0.32% of footways have also `bicycle=designated`, so not much.


> `way[highway=bridleway][bicycle=designated] ` is probably as well although it's much more common (up to 29% of bridleways, https://taginfo.openstreetmap.org/tags/highway=bridleway#combinations) 

I read 3.3% of bridleways have also `bicycle=designated`, so not much.

labels.mss
Florimondable: 0.7% of cases

Phyks:
Indeed, I missed the fact that we could get stats on `bicycle=designated` exactly. That's perfectly fine then!

labels.mss
Phyks: Few examples around https://www.openstreetmap.org/way/380258015, https://www.openstreetmap.org/way/679211980, https://www.openstreetmap.org/way/650712060

For me, this is a tagging error and should not be taken into account. I'll double check.

Phyks: Note that if we consider this case, we should consider the symmetric case also `[oneway != '-1'][oneway_bicycle = '-1']` (any two-way street for cars which is one-way only for bicycles).

Florimondable: Your examples looks like tagging error, though the theoric case can exist in real life.
But not sure if has at least an example in this world.
May be a cycle track in oneway along a bigroad forbiden for cyclist ?

Florimondable: Since the use case is realy rare, we can remove it.

quasart:
Nice ! Many useful path have been enhanced, thank you

PR #198, opened 2019-10-05
Motorcycle-Bicycle parking in mixed blue

### LINKED ISSUE #177: Don't render motorcycle_parking in red
Red should be a color for danger, here with bicyle=yes it should be a more welcome color

Florimondable:
Fix #177 

Fix also a bug with the text not drawn with full opacity whereas the parking circle is transparent cause of limited actess.

Motorcycle-bicycle parking on the left
Capture du 2019-10-05 21-38-38

Text bug
Capture du 2019-10-05 21-42-12
Text fixed
Capture du 2019-10-05 21-37-55

### DIFF
--- a/amenities.mss
+++ b/amenities.mss
@@ -773,6 +773,9 @@
       text-placements: "S,N,E,W,NE,SE,NW,SW";
       text-dy: 6;
       text-dx: 6;
+      [access != ''][access != 'permissive'][access != 'yes'] {
+        text-opacity: 0.33;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


Phyks:
It's a bit hard for me to distinguish the two blue colors for bicycle and shared parkings. Maybe we can enhance a bit the contrast?

Florimondable:
I tried more grey-blue, but in that case it's harder to see shared-parking.
The information is not very important so, it's fine for me.

Phyks:
I'm fine with this for now. We'll anyways have to rework the style and the palette at some point, so this can be postponed.

PR #199, opened 2019-10-05
Render bicycle_road

### LINKED ISSUE #150: Render bicycle_road=yes
See https://taginfo.openstreetmap.org/keys/bicycle_road#overview

Florimondable:
Render bicycle_road as cyclestreet, fix #150 
Fix cyclestreet street colored with their maxspeed.

Before maxspeed bug
Capture du 2019-10-05 23-15-28

After
Capture du 2019-10-05 23-15-37

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -471,8 +471,13 @@ Layer:
             WHEN tags->'maxspeed'='walk' THEN 5
             ELSE NULL
           END AS maxspeed_kmh,
-          bicycle, tags->'motor_vehicle' AS motor_vehicle,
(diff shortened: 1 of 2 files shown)


Phyks:
I merged this one manually. I kept second commit (handle `bicycle_road` the same way as `cyclestreet`) as is.

I swapped the first commit with https://github.com/cyclosm/cyclosm-cartocss-style/commit/cafa1f1908727052d20f1e3b81a634ae3cb23634, which should generate less Mapnik rules (in the ouptut XML file) and the overrides should be easier to understand.

Issue #173, opened 2019-09-25
Distinguish different types of bicycle lanes


westnordost:
In many countries, there are two types of bicycle lanes. Exclusive bicycle lanes and advisory bicycle lanes.

One and a half years ago, I did an extensive research in the quest to establish a tag for that. In a nutshell, it resulted in this tag documentation: https://wiki.openstreetmap.org/wiki/Key:cycleway:lane
The research about the different types of cycle lanes that led to that wiki page: https://westnordost.de/misc/bicycle_lanes.pdf

In my experience, the sharrows (bicycle pictograms on the road) are still better than nothing because car drivers tend to be more respectful of bicycle traffic. (So it would be nice if this was rendered somehow as well.) This might be just a local personal experience, I think the drivers in Hamburg, where I live, are uncertain what the bicycle pictograms mean legally and thus rather drive more careful and considerate. On shared bus lanes, I feel being tailgated by oncoming buses, on sharrows, I don't.

The advisory cycle lanes are to my personal experience actually, even though they give the bicycle a better legal status, only slightly better or about the same as the above. The reason being that they give the bicycles a fixed space, which is usually good, but
- sometimes it is too narrow to be safe (i.e. parking car doors to the right) and car drivers may expect that bicyclists keep strictly within their lanes
- cars often drive or halt on these lanes. Being an own lane, it is more stressful to change into the car lane to overtake the obstacle than if the bicycle pictogram was just printed on the road.

westnordost:
FYI here is an overview what can be tagged with StreetComplete and how the are displayed in the UI in the different countries:

![cyclelanes](https://user-images.githubusercontent.com/4661658/65583126-f0c7a580-df7e-11e9-9e9e-8bcc89bf99c3.png)

Florimondable:
Not sure to understand what feature you're asking :
Would you like to see if a highway with a lane with sharrow to be on the map ?

(For me a cycle lane is always without motor vehicles, but it is because we have not much sharrows in France)

About the street complete image, I wonder how they tag "dashe cycle lane" (for me it's a regular cycle lane) and "suggestion cycle lane".

westnordost:
I am mainly asking for to have a visual distinction for the different values of ways tagged with `cycleway:lane`.

Florimondable:
Ok, so we could render them :
_cycleway=lane + cycleway:lane=advisory_ : the default for cycleway=lane
[...]

PR #200, opened 2019-10-05
Render cycleway=shared_lane

### LINKED ISSUE #148: Render shared_lane?
We currently do not render `cycleway=shared_lane`, such as https://www.openstreetmap.org/way/166351574.

In my opinion, these don't have much value for the cyclists, but might have a bit for the contributors. Not sure whether we should render them (and how)?

Florimondable:
Fix #148 and #173 

Before:
Capture du 2019-10-05 23-32-47

After:
Capture du 2019-10-05 23-32-56

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -471,8 +471,13 @@ Layer:
             WHEN tags->'maxspeed'='walk' THEN 5
             ELSE NULL
           END AS maxspeed_kmh,
-          bicycle, tags->'motor_vehicle' AS motor_vehicle,
-          tags->'cyclestreet' AS cyclestreet,
+          bicycle,
+          tags->'motor_vehicle' AS motor_vehicle,
+          CASE
+            WHEN tags->'cyclestreet' IN ('yes') THEN 'yes'
+            WHEN tags->'bicycle_road' IN ('yes') THEN 'yes'
+            ELSE 'no'
+          END AS cyclestreet,
           CASE
             WHEN oneway IN ('yes', '-1') THEN oneway
             ELSE 'no'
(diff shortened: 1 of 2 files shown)


Phyks:
One remark on the code, we lack the `cycleway=opposite_shared_lane` value which is valid and quite used (most of countraflow bicycle in Paris for instance).

One question (I don't really have an opinion on this), is a `shared_lane` the same quality as a `share_busway`? I would expect the latter to be better since there is only bus and taxi drivers to worry about (limited number of vehicles, professional drivers which should in principle be aware of the bicycles etc).

Florimondable:
There is only 26 cycleway=opposite_shared_lane, and i'm not sure that a countraflow bicycle in Paris is a shared_lane, since there is no lane... So I don't think opposite_shared_lane bring meanings.
If you want to say that the opposite direction of the way have a shared_lane cycleway:left=shared_lane is enough.

I consider third quality infrastructure as "mixed with motor vehicles" whatever they are, the main issue is the volume of traffic either way not the type.

Phyks:
> There is only 26 cycleway=opposite_shared_lane, and i'm not sure that a countraflow bicycle in Paris is a shared_lane, since there is no lane... So I don't think opposite_shared_lane bring meanings.
> If you want to say that the opposite direction of the way have a shared_lane cycleway:left=shared_lane is enough.

Indeed, I was confusing with `cycleway=opposite`, my bad.

> I consider third quality infrastructure as "mixed with motor vehicles" whatever they are, the main issue is the volume of traffic either way not the type.

Current render is fine for me then. We might have to write something about this at some point (maybe in an enhanced legend?). Especially if we want to provide tools for local associations, so that they are aware of the way we consider things.

Phyks:
Merged manually. Matching legend update at https://github.com/cyclosm/cyclosm-website/commit/b2fd5470164ae09a691ba135959962f95ee7f19f.

Issue #194, opened 2019-10-03
Render more landuse like vineyard, orcad, cemetery


Florimondable:
Cemetery seems to be rendered but hard to understand it is a cemetery though they are very usefull to find water in them.

Vineyard https://www.cyclosm.org/#map=14/48.9779/3.2156/cyclosm

Phyks:
Cemeteries are definitely rendered, see https://www.cyclosm.org/#map=16/48.8061/2.3089/cyclosm. This can probably be improved though.

Not sure where is the cemetery on your example, I could not spot it, even with osm.org tiles :)

Florimondable:
«vineyard» not cemetery ;)

Phyks:
Oh, my bad :)

Phyks:
Left to do after #211:

- [ ] `amenity=grave_yard` to be rendered as `landuse=cemetery`
- [ ] `landuse=farmyard` to be rendered as `@farmland`
- [ ] `landuse=allotments` to be rendered as `@farmland`
- [ ] `landuse=retail` to be rendered as `@commercial`
- [ ] `landuse=quarry`
- [ ] `landuse=construction` to be rendered as `@industrial`
- [ ] `landuse=village_green` to be rendered as `@park`
- [ ] `landuse=brownfield` to be rendered as `@industrial`
- [ ] `landuse=recreation_ground` not to be rendered
- [ ] `landuse=garages` not to be rendered
- [ ] `landuse=greenhouse_horticulture` to be rendered as `@farmland`
- [ ] `landuse=greenfield` to be rendered as `@heath`
- [ ] `landuse=railway` not to be rendered
- [ ] `landuse=landfill` to be rendered as `@industrial`
- [ ] `landuse=plant_nursery` to be rendered as `@farmland`
- [ ] `landuse=salt_pond` to be rendered as `@heath`

What do you think @Florimondable?

Florimondable:
landuse=farmyard to be rendered as @farmland : disagree, something else, it mostly buildings, house and a farm yard, no crop here.
landuse=allotments to be rendered as @farmland : more like a garden or village_green
landuse=plant_nursery to be rendered as @farmland : like meadow

Phyks:
Done, except `landuse=farmyard` which is already good looking in my opinion.

PR #211, opened 2019-10-08
Add landuse vineyard and orchard, add symbol for cemetery

Florimondable:
Link to #194, but amenity=grave_yard is not rendered.
Many other landuse are still waiting to be added.

Capture du 2019-10-08 22-02-35
Capture du 2019-10-08 22-03-30
Capture du 2019-10-08 22-03-12


Phyks:
Merged

Issue #226, opened 2019-10-25
secured parking icon proposition


quasart:
Hello everybody,
I like the little roof you've just added on the covered bicycle parking.

I think secured parking icons could follow the same idea : the blue mark could be "locked" inside a little square. Here is a preview/comparison, first line is current, bottom line is what I have in mind
![icons](https://user-images.githubusercontent.com/3080387/67560876-0d8cff80-f71c-11e9-9611-62e0668f250a.png)

This looks more understandable and more consistent to me.

Florimondable:
I like the idea

PR #228, opened 2019-10-29
secured bicycle parking icon rework

quasart:
Hello,
here is a rework of icon files for #226.

### DIFF
--- a/symbols/amenity/bicycle_motorcycle_parking_matrix.svg
+++ b/symbols/amenity/bicycle_motorcycle_parking_matrix.svg
@@ -9,13 +9,13 @@
    xmlns="http://www.w3.org/2000/svg"
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   width="210mm"
-   height="297mm"
-   viewBox="0 0 210 297"
+   width="640"
+   height="128"
+   viewBox="0 0 640 128"
    version="1.1"
    id="svg8"
-   inkscape:version="0.92.4 5da689c313, 2019-01-14"
-   sodipodi:docname="bicycle_parking_full.svg">
+   inkscape:version="0.92.3 (2405546, 2018-03-11)"
+   sodipodi:docname="bicycle_motorcycle_parking_matrix.svg">
   <defs
      id="defs2" />
   <sodipodi:namedview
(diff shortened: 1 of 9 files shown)


Florimondable:
Nice !
The roof of the secured parking icons (squared ones) doesn’t have the same width as the other ones.
Can you correct that please ?

Can you make a snapshot of an example of the map with these new icons ?

Phyks:
I also have a concern with limited access bicycle parkings. We currently render them with an opacity of 0.33 and this does not render very well with these new icons. Not sure how to fix this, maybe increase the size of the white inner ring?

Screenshot_2019-10-30 CyclOSM — Kosmtik

This is located at https://www.cyclosm.org/#map=20/48.85306/2.36761/cyclosm.

quasart:
Hello,
Opacity looks applied on every subpart of the svg, and not on the icon itself. I’ll try to provide « flat » svg without overlapping shapes. It should look better.

quasart:
I've changed the square into simple contour corners. I hope it has good rendering, but unfortunately I have no way to test it for now.

Phyks:
Sorry for coming back late on this :/ Opacity looks good now, thanks!

Screenshot_2019-11-06 CyclOSM — Kosmtik

One minor issue, in my opinion, is that the "lock" is not strong enough. 

2Screenshot_2019-11-06 CyclOSM — Kosmtik

The only main visual difference once rendered are the small triangles in the corners.

What do you think?

For the record, the bicycle parking icons are rendered with a size of 8 pixels (could be increased a bit to 10 if required, but render is currently the same).

Florimondable:
> For the record, the bicycle parking icons are rendered with a size of 8 pixels (could be increased a bit to 10 if required, but render is currently the same).

I try it myself, and I wasn’t happy with the result.
So I try to put the dot into a full square box, and I figure out that the icon is size fixed as you mentioned.
But I don’t like the idea to add an other rule for just the size of an icon.
May be we can try to reduce the size of the dot, or render the dot in white and the "house" in blue ?

Phyks:
> I figure out that the icon is size fixed as you mentioned.

Current icon size is 8 pixels (width, height is determined from the marker size). We could enlarge it to 10 pixels, but it does not change much :/

quasart:
Hello,
if width is fixed but not height, we may use height to make secured icon stronger. Here is a new try with a thicker base added :
test

Phyks:
> if width is fixed but not height

It should be the case, and indeed it should be fine to have non-square icons.

> Here is a new try with a thicker base added :

Looks better! Maybe we should make the top stronger as well?

2Screenshot_2019-11-06 CyclOSM — Kosmtik

quasart:
The top of the « covered » icons ? I kept the uncovered one light to make clear difference

Phyks:
> The top of the « covered » icons ?

Yes, only for the covered ones (actually,the locked ones). Currently, the locked ones have a strong basis and a lighter top. What about locking symmetrically?

quasart:
Here I tried to make top of lock stronger for both covered and uncovered secured

Phyks:
Thanks!

Florimondable:
Result for the record :
Capture du 2019-11-23 21-52-29

Issue #197, opened 2019-10-04
Decrease shield road (and route)


Florimondable:
https://www.cyclosm.org/#map=11/49.6970/4.1968/cyclosm
too big, too colorfull, only big roads, minor roads could have shield also

Florimondable:
can be done also on cycle routes in high zoom, there shields are much more visible than the transparent route.

Florimondable:
For roads I think we should remove shields, they are ok when you want to follow main road but on a bike it is the opposite you prefer to avoid them, so it is better to render road ref for more roads but smaller. So just render the ref along the road as do IGN or OTM for instance.
Here a try :
![Capture d’écran de 2019-11-01 22-26-14](https://user-images.githubusercontent.com/47089717/68057997-06578a00-fcf8-11e9-9a47-a5bf81bffe78.png)

I think we should also render ref/name of track/path, they can be usefull for people on MTB/Trekking/Tourism.

I also found that there is some cleaning to do in the road name layer.

PR #233, opened 2019-11-01
Render road ref along them, smaller cycle route shield, cleaning

Florimondable:
link to #197 

cyclosm_road_route_ref_city
cyclosm_road_names_country


labels.mss
```diff
@@ -129,166 +129,103 @@
...
+      text-clip: false;
+      text-placement: line;
+      text-face-name: @sans;
+      text-repeat-distance: @major-highway-text-repeat-distance;
+      text-halo-radius: @standard-halo-radius;
   }
 }
 
 #roads-text-ref {
-  [highway = 'motorway_trunk'],
```
Phyks: The two of them were merged https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/project.mml#L2011 in SQL queries. We should either keep `motorway_trunk` or drop the merge in `roads-text-ref` SQL.

labels.mss
```diff
@@ -129,166 +129,103 @@
...
+      text-size: 12;
     }
   }
 }
 
 #roads-text-name {
   [highway = 'motorway'],
   [highway = 'trunk'],
-  [highway = 'primary'],
-  [highway = 'construction'][construction = 'motorway'],
```
Phyks: Why dropping the name of roads under construction?

labels.mss
```diff
@@ -629,6 +521,7 @@
   shield-face-name: @sans;
   shield-clip: @shield-clip;
   shield-file: url("symbols/shields/[type]_[width]x[height].svg");
+  shield-transform: "scale(0.75)";
```
Phyks: I'd rather edit the script https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/scripts/generate_shields.py to generate shield the right size. I'll handle this after this PR has been merged.

project.mml
```diff
@@ -2101,20 +2100,16 @@ Layer:
...
           ('tertiary_link', 200),
-          ('service', 150),
-          ('cycleway', 140),
-          ('path', 130),
-          ('construction', 10)
+          ('service', 150)
         ) AS ordertable (highway, prio)
         USING (highway)
         WHERE highway IN ('motorway', 'motorway_link', 'trunk', 'trunk_link', 'primary', 'primary_link', 'secondary', 'secondary_link', 'tertiary',
-          'tertiary_link', 'residential', 'unclassified', 'road', 'service', 'pedestrian', 'raceway', 'living_street', 'construction', 'cycleway', 'path')
```
Phyks: With these edits, we lost the names of raceways, is it expected?

Phyks:
:+1:

We have a bit of readability issues at low zooms (city or less).

https://cyclosm.org/#13/48.8779/2.2867

Screenshot_2019-11-08 CyclOSM — Kosmtik

https://cyclosm.org/#11/48.8401/2.0098

Screenshot_2019-11-08 CyclOSM — Kosmtik 2

labels.mss
Florimondable: What’s the point of having them ?

labels.mss
Florimondable: I tried, I failed 😭

project.mml
Florimondable: Yes, I don’t see the point of having raceway names in a pov of a cyclist.

Florimondable:
> We have a bit of readability issues at low zooms (city or less).
May be, not sure, +1 to the font size ?

labels.mss
Florimondable: I don’t remind to have done that change, must come from the copy past I have done to write that layer from name layer.

labels.mss
Phyks: We'd lose the name on "Avenue de la République" here for instance https://www.cyclosm.org/#map=19/48.80997/2.30184/cyclosm.

Florimondable: The road is not even rendered, the name is on nothing, so it is a good fix :]

Issue #138, opened 2019-06-19
Render road with limited access to bicycle


Florimondable:
For instance https://www.openstreetmap.org/way/5213104#map=18/48.84117/2.32132

Florimondable:
An other example in Pere Lachaise, Paris, a cemetery forbidden for cyclist but service roads rendered as blue due to 20kmh maxspeed.
https://www.cyclosm.org/#map=17/48.86104/2.39497/cyclosm

Idea:
- fade the color ?
- render x on the road ?

Phyks:
> use yellow/orange/red ?

Should not use this in my opinion, way to visible for something we want to avoid (same issue with current primary / secondary).

> render x on the road ?

Having an "- x - x -" pattern on the outline of the way might be nice, no?

Florimondable:
I think render the road in grey can do the job, it already does well for motorway.
Grey doesn't make the road visible and make it think as dead road.
And it's easy to do.

Phyks:
Is grey itself enough? :+1: for having it grey, but there should maybe be something else to make it clear this is not only discouraged but forbidden as well.

Phyks:
See https://github.com/cyclosm/cyclosm-cartocss-style/issues/225

Phyks:
Mostly fixed by #236.

Florimondable:
a map with cross on limited access road is visible with the MTB OSM map http://mtbmap.cz/#zoom=18&lat=45.927382&lon=6.148454

PR #236, opened 2019-11-03
Fill in grey road without bicycle access

Florimondable:
fix most of #138 
May be we need to fix the outline of main road and path (too faded) and some other tweaking, but overall the render is a good improvement imho.

cyclosm_nobicycle_auteuil
cyclosm_nobicycle_perelachaise
cyclosm_montparnasse_nobicycle

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1417,7 +1417,7 @@ come in as well.
           line-color: @speed20-fill;
       }
       [maxspeed_kmh < 10],
-      [motor_vehicle='no'][bicycle!='no'] {
+      [motor_vehicle='no'][can_bicycle!='no'] {
           line-color: @speedWalk-fill;
       }
     [can_bicycle='no'] {
--- a/roads.mss
+++ b/roads.mss
@@ -1361,6 +1361,7 @@ come in as well.
 #bridge::inline[zoom>=11] {
   /* -- colors & styles -- */
   line-color: @standard-fill;
+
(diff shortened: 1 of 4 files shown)


roads.mss
```diff
@@ -1765,9 +1771,9 @@ come in as well.
       [type='bridleway'][can_bicycle='yes'] {
         line-color: lighten(@path-fill,@surfaceLighter1);
       }
-      [type='path'][can_bicycle='no'] {
-        line-color: lighten(@path-nobicycle,@surfaceLighter1);
-      }
+     /* [type='path'][can_bicycle='no'] {
+        line-color: lighten(@standard-nobicycle,@surfaceLighter1);
+      }*/
```
Phyks: We can remove this dead code :)

roads.mss
```diff
@@ -1419,16 +1417,24 @@ come in as well.
           line-color: @speed20-fill;
       }
       [maxspeed_kmh < 10],
-      [motor_vehicle='no'][bicycle!='no'] {
+      [motor_vehicle='no'][can_bicycle!='no'] {
```
Phyks: I was using `bicycle` here to only check if cycling was allowed (and not general access tags etc). I think this is the only use of `bicycle` in this file, so we can probably remove it from SQL queries.

Phyks:
Should have been merged with my two remarks.

Issue #263, opened 2019-12-16
Tourism point of view labels are still brown


Florimondable:
exemple https://www.cyclosm.org/#map=20/48.87328/2.33221/cyclosm

Florimondable:
would be fixed by #273

Issue #259, opened 2019-12-02
Reduce visibility of access=private


jodastephen:
In my area, I've added quite a lot of service roads (`highway=service`) and rear access tracks (`highway=track`) with `access=private`. However, the current style ignores the access level and thus highlights these as highly visible (completely the opposite to your goals).

Here is an example of a `highway=service, access=private`: [link](https://www.cyclosm.org/#map=19/51.41166/-0.23450/cyclosm)

Here is an example of lots of `highway=track, access=private`: [link](https://www.cyclosm.org/#map=16/51.3940/-0.2170/cyclosm)

Can I request that routes with `access=private` are not highlighted in this way? Thanks

Phyks:
Indeed, we could probably reduce their opacity, similary to what we do for amenities.

Florimondable:
> Here is an example of lots of `highway=track, access=private`: [link](https://www.cyclosm.org/#map=16/51.3940/-0.2170/cyclosm)

For me this is bad tagging, track is for agriculture or forest purpose road, here it's a service road like alley or driveway, or path if very narrow.
https://wiki.openstreetmap.org/wiki/Tag:service%3Dalley
https://wiki.openstreetmap.org/wiki/Tag:service%3Ddriveway

> 
> Can I request that routes with `access=private` are not highlighted in this way? Thanks
Yes we can work on that, we already did road with forbidden access to bicycle in grey, may be the same here ?

Florimondable:
Be aware that we already fade color for road in tunnels.

Florimondable:
I'm again ok with the Idea: just render with the same color as bicycle=no roads.
Because the result for a cyclist is the same "you can't ride there".

The only problem I can see here is that we don't know if you can walk on the road or not, but already have the issue with bicycle=no color. May be we could have an other color (more white) for road open for walking but not for cycling ?

jodastephen:
My opinion is that anything tagged with `access=private` is indicating no access for wlaking or cycling - it implies `foot=private` and `bicycle=private` in the absence of specific `foot` or `bicycle` tags. As such, I think anything with `access=private` should be "greyed out".

FWIW, the same is true of `access=no` - it implies `foot=no` and `bicycle=no`.

Florimondable:
Yes with d6e66a79aa9e80c5705e7cd29f44c5b8846b90d8
bicycle(or access if no bicycle tag)=no/private will be rendered in grey :)

Issue #262, opened 2019-12-09
Train station is too much visible at z12


Florimondable:
Paris looks like to have varicella https://www.cyclosm.org/#map=12/48.8593/2.3253/cyclosm
Subway stations can be removed
Train stations could be smaller.

Phyks:
Problem is any solution to this should also make the train stations highly visible in the country side: https://www.cyclosm.org/#map=13/48.7559/1.4651/cyclosm.

Concerning subway stations, I guess we should exclude `station=subway` (https://wiki.openstreetmap.org/wiki/Key:station?uselang=fr) unless there is an explicit bicycle access? Would that work globally (Strasbourg where bicycles are allowed in tram stations or San Francisco for instance)?

Last idea, which could work for POIs as well, could be to count objects and only return them at given zoom level if there are less than a given number. Typically, Z12 in Paris, a lot of stations > do not display (until Z13 or 14 where the density is much lower). Z12 in the countryside, only a few stations, display them.

Florimondable:
My PR do the job well
@Florimondable
Render subway station from z14 instead of Z12, to avoid too much stat…
51f8578
@Florimondable
Render halt train station.
2e5f9fc

Issue #255, opened 2019-11-26
In middle zoom (Z9-10) we barely see roads, specialy in the Alps


Florimondable:
I’d like to be able to find roads in the Alps but I barely can see them in middle zoom which is the good zooms to find where to go through the mountains.
This can aslo apply for plane.

Could take example on OpenTopoMap
https://www.cyclosm.org/#map=10/44.1620/7.0113/cyclosm

This should apply only for motorway/primary/secondary/tertiary

Z11 coul be also rework to have roads a little more visible.

Florimondable:
link to #183

Florimondable:
Can you let it open, I'm not sure it will be completely closed after #273

Phyks:
Is this one still active (especially since we added contours back?)?

Florimondable:
Contours is not present à z 9-10 ;)

I think the issue is still there, but probably not easily solvable.

PR #273, opened 2020-01-03
Polishing 3 : railway stations, roads at z9-10, access private

### LINKED ISSUE #272: Halt train station are not rendered
https://wiki.openstreetmap.org/wiki/Tag:railway%3Dhalt
There widely used in Germany for instance, but also in France.

Florimondable:
Code cleaning.
fix #272 and #262 
fix almost #255 but I'm not able to render tertiary (WTF?)

Capture du 2020-01-03 23-11-53


Florimondable:
Add allotments, wetland and some naturals (glacier is still missing at least) (for #229)
Should fix #263 (tourism point of view in brown) and #265 (allotments)

Florimondable:
fix #259 private access roads in grey

access_private_montrouge

roads.mss
```diff
@@ -1783,7 +1783,10 @@ come in as well.
 
     /* Surface ok for treking/gravel/cyclocross/city bike (28mm < tyres <= 40mm). */
     [surface_type='cyclocross'] {
-      line-dasharray: 6,12;
+      line-dasharray: 4,8;
```
Phyks: Should we update the legend?

Florimondable: no it's fine

roads.mss
```diff
@@ -38,35 +38,41 @@ as well. */
...
-    }
-  }
+  [type='motorway_link'],
   [type='trunk'],
   [type='trunk_link'] {
-      line-color: @motorway-trunk-cycle-fill;
-      [bicycle='no'] {
-        line-color: @motorway-trunk-fill;
-      }
+      line-color: @motorway-trunk-fill;
```
Phyks: Why are we discarding the `bicycle` switch? motorway/motorway_link should not be accessible by bikes by default, but trunk/trunk_link should be: https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions#Default.

roads.mss
```diff
@@ -38,35 +38,41 @@ as well. */
...
   }
-  [type='primary'] { line-color: @primary-line; }
-  [type='secondary'] { line-color: @secondary-line; }
-  [type='tertiary'] { line-color: @standard-line; }
+  [type='primary'] {
```
Phyks: We should probably drop `primary-line`, `secondary-line` and `standard-line` from road colors, they are unused after this change.

project.mml
```diff
@@ -720,7 +720,7 @@ Layer:
           highway AS type,
           bicycle
```
Phyks: Bicycle no longer used in the cartocss code, we could remove it from SQL query.

Phyks:
> but I'm not able to render tertiary (WTF?)

`roads_med` layer is using `planet_osm_roads` table (not `planet_osm_line` table as used on upper zoom levels). This table only contains a subset of roads, which are useful at low zoom level (for perf), see https://gis.stackexchange.com/questions/37099/in-osm2pgsql-how-is-the-planet-osm-roads-table-populated.

In particular, it only has very few tertiary. The logic is basically described in https://github.com/openstreetmap/osm2pgsql/blob/master/style.lua (we are not exactly using this, but the core ideas of the logic are the same, check for the `roads` items). The few tertiaries listed in `planet_osm_roads` table are coming from artifacts from other items being added into this table (administrative boundaries along a tertiary for instance).

I'd expect the `tertiary` to have too much clutter at Z9/Z10, I would be in favor or removing them from  `roads_med` layer.

Phyks:
Few comments above, once settled I'm ready to merge this PR and publish a new version of the CyclOSM style :)

roads.mss
Florimondable: because the test is not correct and different than higher zoom, we just test bicycle, where we should test more, a can_bicycle should be added otherwize

Issue #270, opened 2019-12-30
Military area not visible into forest


Florimondable:
> Military features may be missing on some zoom levels: https://www.cyclosm.org/#map=13/51.8260/8.7798/cyclosm. The entire area is military and non-accessible.

PR #285, opened 2020-01-18
Make military areas visible even when both military and landuse are set

hiddewie:
Fixes https://github.com/cyclosm/cyclosm-cartocss-style/issues/270

The query missed military areas where both `military=...` and `landuse=military` are set. For example a military range with `landuse=military` and `military=range`. This feature would be picked up by the query, but not be styled on the map.

Example query for such data: https://overpass-turbo.eu/s/PSG

Before:

Annotation 2020-01-18 213654

After

Annotation 2020-01-18 223200

### DIFF
--- a/base.mss
+++ b/base.mss
@@ -138,7 +138,7 @@
   }
 }
 
-#protected-areas[zoom >=7 ] {
+#protected-areas[zoom >= 7] {
   line-color: darken(@wooded,25%);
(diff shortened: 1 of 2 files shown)


Phyks:
Thanks for this PR! and sorry for the delay for the merge :/

hiddewie:
Great, thanks!

Issue #276, opened 2020-01-09
Rework regional/local routes at mid zoom level (10/11)


Phyks:
> You mention above that it is more useful to see a regional route from a
> small zoom level in France than in The Netherlands. I agree.
> 
> The discussion comes down to "do I want to see a regional cycling route on
> a small zoom level" which translates to "what does a regional cycling route
> mean (compared to national or international)". In
> https://wiki.openstreetmap.org/wiki/Relation:route#Cycle_routes_.28also_mountain_bike.29
> it
> says that regional routes cross regions. That is true for all cycling
> routes on the map so the data seems to be correct.
> What I noticed is that on
> https://wiki.openstreetmap.org/wiki/Tag:network%3Drcn says the relation can
> have *network:type=node_network* when a route is part of a node network. A
> Dutch national route (https://www.openstreetmap.org/relation/3339) does not
> have a node network attribute, but
> https://www.openstreetmap.org/relation/4221186#map=12/52.2209/6.9567 has
> *network:type=node_network*.
> 
> If I would have to make a choice, I would experiment with starting to show
> regional routes starting on zoom level 10. At least in the Netherlands they
> are useful from that level.
> Maybe it would be possible to use network:type, and then more fine-grained
> styling is possible.
> 
> Finally another cycling map (although it is really ugly):
> https://cycling.waymarkedtrails.org/#?map=9!52.1807!6.5262. I do agree with
> the zooming choices in the Netherlands.

Phyks:
Closed by #286

PR #286, opened 2020-01-19
Do not show regional cycling network routes on zoom 8-10

hiddewie:
This PR improves the cycling network visualization in The Netherlands and Belgium for any regional routes marked with `network:type = 'node_network'` (impact, see https://taginfo.openstreetmap.org/tags/network%3Atype=node_network#map)

Large improvement for issue #276.

- Don't include cycling network regional routes (node network type) for display on zoom levels 8 to 10
- Don't show cycling network regional routes (node network type) labels on any zoom level (the route names are not important, the nodes are).

## Zoom 8

Old 

image

New 

image

## Zoom 10

Old 

image

New 

image

## Zoom 11

Unchanged (except for the labels)

image

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -954,6 +954,7 @@ Layer:
         FROM planet_osm_line
         WHERE (route='bicycle' OR route='mtb')
           AND tags->'network' IN ('icn', 'ncn', 'rcn')
+          AND NOT (tags->'network' = 'rcn' AND tags->'network:type' = 'node_network')
         ORDER BY CASE
           WHEN tags->'network' = 'icn' THEN 0
           WHEN tags->'network' = 'ncn' THEN 1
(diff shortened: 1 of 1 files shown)


Florimondable:
Hi, thank you for pull request.

This is a difficult subject since netherland is the only country to have a such dense cycle route network.
If we remove regional cycle route from z8-10 we remove from the map big routes that go through regions and where low zoom visualiszation is needed.
Example in France https://www.cyclosm.org/#map=9/48.0248/-1.3719/cyclosm

I realy wonder if the those route should not be lcn.

Florimondable:
>     * Don't show cycling network regional routes labels on any zoom level (the route names are not important, the nodes are).

I'm against this change since most of the country don't use node network.
And route names can be still usefull.

hiddewie:
@Florimondable Sorry, the wording was not correct in the description. 

This PR only changes the case for `network:type = 'node_network'`. This should only be the case in The Netherlands, Belgium (and appenently somewhere in the north of Germany). See https://taginfo.openstreetmap.org/tags/network%3Atype=node_network#map

All other (normal) regional networks should be rendered the same way as before (routes and labels). 

I've corrected the PR description.

Phyks:
Really nice idea to rely on the `network:type = 'node_network'` tag! Thanks a lot for this PR!

Florimondable:
@hiddewie I think we have a regression #297 
But I can't see why

Issue #269, opened 2019-12-30
Cycle junction node hidden by river name


Phyks:
See https://www.cyclosm.org/#map=16/52.3387/6.5977/cyclosm and https://www.cyclosm.org/#map=17/52.33878/6.59575/cyclosm.

PR #287, opened 2020-01-19
Make cycle network nodes overlap and increase size on small zoom levels

hiddewie:
Should solve #269

- Make cycle network nodes render before place names and waterway names.
- Add placement of place names with simple placement multiple options with dx=25 and dy=20. 
- Makes the background of the node background/halo a bit larger for zoom 13 to 15. This makes the nodes much more prominent for planning a route.

Problem area with cycle network node overlapping with water name:

Old

image

New

image

Overview of cycling network area

Old

image

New

image


Phyks:
:+1: for increasing the size on small zoom levels. Overlapping renders very well for the river issue, but the latest screenshot shows issues with city names (Rijssen or Almelo for instance).

Is this issue of hidden cycle node mainly coming from bad interaction with river names or are there also other sources of conflict (e.g. city names) in the current code?

Did you give a try with something like  `shield-placements`? The position of the node will be slightly off but it might fix the overlap conflict.

Last question is whether we should make the city names the topmost layer?

hiddewie:
Hmm that is an interesting problem (overlapping place names and node numbers). I am still unsure of the allow overlap.

I think the issue is bigger than just the original river name overlapping with the node number. For example in the overview image in the current code, the node numbers within the city of Rijssen are not visible at all. 

It would be nice if the place name could 'move' until it finds a place without overlap with existing items on the map. My knowledge of the carto CSS is not that good.

hiddewie:
I just found a new option for most symbolizers: `ignorePlacement`. It is allowing overlap, but then for layers *after* it.

See https://github.com/mapnik/mapnik/wiki/PointSymbolizer#configuration-options

Phyks:
Hi @hiddewie, sorry for coming back late on this... This sounds interesting!

Did you give it a try? Otherwise, I'll try to test it and see what we can do.

Hope everything is well for you, best,

hiddewie:
Hello @Phyks.

All is fine! I did test the `ignorePlacement`, but I could not get it to work. It might be my versions of Carto CSS or something else. 

For this PR: either the `ignorePlacement` must be made to work, or I should close this PR I think. It is not a too large issue, and my initial solution created more problems than solutions in certain places/zoom levels.

Phyks:
I gave it a try and it seems `ignorePlacement` is only available for `PointSymbolizer` and `MarkerSymbolizer` (while we use `ShieldSymbolizer` here for the cycle junction nodes). I see 3 possible solutions so far:

1. Moving up the `cycle-junction-node` layer before the text layers in the `project.mml` and using `text-allow-overlap: true` on the place names. It works, I did not spot mistakes but I don't like much the idea of having placenames with allow overlap.

2. Same as 1, but using `text-placements` on the placenames instead of allowing overlap. 



3. Finding a more elaborate way. This could be finding a way to drop the shield and only rely on point/markers to be able to use `ignorePlacement`.

Solution 2 is implemented in https://github.com/cyclosm/cyclosm-cartocss-style/commit/88e6b8ebd16f9c615e33d18af6824882de7022d0. I think it could be fine with a bit more of fine tuning (`text-dx` / `text-dy` is too large here I think).

The downside of solution 2 is that it is initially meant for labels of a marker, so `dx`/`dy` is fixed (and it does not mean "put in wherever you want in this enveloppe"). We could refine it furher by moving from a point geometry for placenames to a polygon geometry, but this seems to be possibly quite a lot of work and/or extra computation.

Phyks:
Note that we can apparently use X for exact position in placements list. https://github.com/mapnik/mapnik/wiki/TextSymbolizer#simple

Seems to be ideal for solution 2.

hiddewie:
I agree that option 1 is not desirable, because at a later stage other symbols might be added to the map on certain zoom levels, that will suddenly generate overlap with placenames.

Option 2 seems good! The `dy` of `15px` seems like a large value. Why not have `X,N,S,E,W,NE,SE,NW,SW,<size1>,<size2>`, etc with some font sizes? That way, first the large place name is tried in all places, then the smaller. If nothing else works it is not rendered, but zooming in (or out) will still show the placename. (I have to try this locally)

Option 3 seems too difficult for the problem.

hiddewie:
OK I tried your implementation of option 2, and then tweaked the values a bit. The end result looks satisfactory, and seems to solve multiple problems with the rendering of cycle junction nodes.

Some screenshots.

### http://0.0.0.0:6789/cyclosm/#12/52.3038/6.5132 

Rijssen, problematic with 2 cycle nodes practically on top of each other and on top of the place name

(zoom 12, unchanged)

image

(zoom 13, shows 1 cycle node because the other overlaps, and placename)

image

(zoom 14, shows 2 cycle nodes and the placename with moved to E placement because of station name)

image

(zoom 15, shows 2 cycle nodes and the placename, lot's of space available)

image

### http://0.0.0.0:6789/cyclosm/#15/52.3360/6.5974

Waterway label overlaps with cycle node

(zoom 14)

image


(zoom 15)

image

(zoom 16, shows both the cycle node and river name)

image


### Other examples

http://0.0.0.0:6789/cyclosm/#13/52.3552/6.6404

image

http://0.0.0.0:6789/cyclosm/#13/52.2471/6.8483

See the many more rendered nodes in the middle of Hengelo (nr 97) and Enschede (nr 60), and Usselo (nr 35).

image

labels.mss
```diff
@@ -1,6 +1,7 @@
 /* LABELS.MSS CONTENTS:
- * - area labels
- * - waterway labels
+ * - Area labels
+ * - Waterway labels
+ * - Bycicle route labels
```
Phyks, line 4: Bicycle (reversed y and i)

placenames.mss
```diff
@@ -307,6 +319,10 @@
     text-wrap-width: 70; // 5.0 em
     text-line-spacing: -0.70; // -0.05 em
     text-margin: 9.8; // 0.7 em
+    text-placement-type: simple;
+    text-placements: "N,S,E,W,NW,NE,SE";
```
Phyks, line 323: Did you try adding the "X" position as well?

Phyks:
This looks really good, thanks for the update and detailed diff! Two minor comments, but I'm ok with merging it!

cc @Florimondable for eventually some additional feedback

placenames.mss
hiddewie: Yes, I did, but this screwed up the placement. As far as I was able to determine, if there is an X in the placement list it will try placement on its center and then abandon placement if there is overlap. It will not try N, S, E, W, etc...

But this was hard to test well and there is little to no documentation from Mapnik.

Phyks: Indeed. As far as I understand, the issue is actually with `text-dx`/`text-dy` which apply including in the X placement case... and Mapnik doc is really scarce on this.

Digging into Mapnik code to try to have a better understanding, I discovered the "C" placement which is totally undocumented and seems to do magic.

Here is the current view in CyclOSM:
old

And the one with the leading "C" placement:
new

It seems to have both benefits from your changes and stricter positionning (look at Nijverdal position, West, which is almost unaffected by these changes).

So, I merged your PR adding this extra commit https://github.com/cyclosm/cyclosm-cartocss-style/commit/a50958ac5ffab263af67d25944e3ee6019195209. I'm welcoming feedbacks if you think I missed something with the C placement.

hiddewie: Very well, thanks for the addition. Indeed I did not find `C` placement.

(Somewhere deep on my list is to test and improve the Mapnik 3 documentation. That would benefit map creators, including me, a lot)

PR #299, opened 2020-02-06
Fix null in 'not' in cycle route node query for zoom 8-10

### LINKED ISSUE #297: Can't see anymore regional route at z8-10
Probably a regression after #286

exemple :
https://www.cyclosm.org/#map=10/48.1083/-1.2051/cyclosm
https://cycling.waymarkedtrails.org/#route?id=1019787

hiddewie:
Fix #297.

There were null values in the query because not all ways have the tags in the query (of course...). And `null` is not `false` in SQL, but resolves the whole WHERE to `null`.

### Current live map

Zoom 8

image

Zoom 10

image

### After fix

Zoom 8

image

Zoom 10 

image

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -977,7 +977,7 @@ Layer:
         FROM planet_osm_line
         WHERE (route='bicycle' OR route='mtb')
           AND tags->'network' IN ('icn', 'ncn', 'rcn')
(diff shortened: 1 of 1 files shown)


Phyks:
Nice catch! Thanks!

PR #300, opened 2020-02-08
Add cinema, theatre, library, casino, bureau de change

### LINKED ISSUE #296: render movie theatre
No movie theatre on the map ? :(

Florimondable:
fix #296


amenities.mss
```diff
@@ -321,6 +321,41 @@
     marker-clip: false;
   }
 
+  [feature = 'amenity_cinema'][zoom >= 17] {
```
Phyks: I would only show these extra POIs from zoom 18. What do you think?

amenities.mss
```diff
@@ -1075,21 +1117,6 @@
     text-placement: interior;
   }
 
-  [feature = 'tourism_information'][zoom >= 19],
```
Phyks: Why changing this render?

amenities.mss
```diff
@@ -1004,19 +1039,26 @@
     text-dx: 12;
   }
 
+  [feature = 'tourism_information'][zoom >= 19],
+  [feature = 'tourism_information']["information"='office'][zoom >= 17],
   [feature = 'tourism_artwork'][zoom >= 19],
   [feature = 'tourism_gallery'][zoom >= 17],
-  [feature = 'historic_city_gate'][zoom >= 17],
   [feature = 'tourism_museum'][zoom >= 17],
```
Phyks: Why changing the render of these `man_made` / `tourism`?

amenities.mss
Florimondable: z17 because most of the usual amenities (shops...) begin at z17

amenities.mss
Florimondable: I just reordered the list to have a to have the same kind of POI together, and changed @memorial to @culture because @memorial has no sense here.
Final rendering is the same.

amenities.mss
Florimondable: Useless differences from the others

amenities.mss
Phyks: I would rather have them show only from Z18 as the map is already very heavily loaded at Z17 and these amenities are quite useless for a bike (except sometimes for the final destination).

Florimondable: heavily loaded in Paris because it’s Paris, but in Bourgoins Jailleu not that much ;)
https://www.cyclosm.org/#map=17/45.58638/5.27592/cyclosm

Issue #253, opened 2019-11-25
Cycle routes in low zooms are noisy


Florimondable:
The lines are not streight where they should :

https://www.cyclosm.org/#map=7/48.261/-0.791/cyclosm

![image](https://user-images.githubusercontent.com/47089717/69542431-b9c83d00-0f8b-11ea-802f-985f3ff9a426.png)

PR #305, opened 2020-02-11
Try to fix noise bicycle route at low zooms

Florimondable:
to #253 

First commit use smaller grid to avoid its bad effect of filling the grid even if there is only one node in it Which result in noisy pixels around the route and like the opposite effect of antialiasing, that make the route more aliased and bigger. Results are ok.

Second commit use of ST_Symplify in order to realy simply the geometry which result with less nodes on the ways and hence better line rendering not aliased. Good result.

_Performance has not been tested, I don't know how to do that._

Before:
Capture du 2020-02-11 23-43-15_snap4

SnapToGrid after first commit, bigger grid (256):
Capture du 2020-02-11 23-44-00_snap256

Symplify:
Capture du 2020-02-11 23-42-02_simplify4

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -936,7 +936,7 @@ Layer:
     <<: *osm2pgsql
     table: |-
       (
-        SELECT ST_Snaptogrid(way, !pixel_width! / 4) AS way, route, tags->'network' AS type, tags->'state' AS state
+        SELECT ST_Snaptogrid(way, !pixel_width! / 1024) AS way, route, tags->'network' AS type, tags->'state' AS state
         FROM planet_osm_line
         WHERE (route='bicycle' OR route='mtb')
           AND tags->'network'='icn'
@@ -953,7 +953,7 @@ Layer:
     <<: *osm2pgsql
     table: |-
       (
-        SELECT ST_Snaptogrid(way, !pixel_width! / 4) AS way, route, tags->'network' AS type, tags->'state' AS state
+        SELECT ST_Snaptogrid(way, !pixel_width! / 256) AS way, route, tags->'network' AS type, tags->'state' AS state
         FROM planet_osm_line
         WHERE (route='bicycle' OR route='mtb')
           AND tags->'network' IN ('icn', 'ncn')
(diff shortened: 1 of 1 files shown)


Florimondable:
using tuning_sql.py, I got those results
z3, no optim (just way) : 352ms, 120 lines
z8, no optim (just way) : 112ms, 442 lines
z3, snaptogrid/4 : 1076ms, 11 lines
z8, snaptogrid/4 : 97ms, 261 lines
z3, simplify : 341ms, 120 lines
z8, simplify : 92ms, 442 lines

So snaptogrid is lower in low zooms (z3) than no optim, a little faster in higher zoom (z8) than no optim.
Simplify is faster than snaptogrid and no optim in low zooms (z3), and slightly faster in higher zoom (z8). Simplify has the same number of result lines since it doesn't remove any line, but it simplifies them a lot.

A render test could be done.

Florimondable:
render time with
`python ./scripts/render_single_tile.py -c 3 48.841 2.444 -s mapnik_prod.xml -t`
script found in https://github.com/giggls/openstreetmap-carto-de/blob/master/scripts/render_single_tile.py
ST_Snaptogrid: (z3-z7-z8-z10)
rendering time: 1.164444 seconds
rendering time: 2.836520 seconds
rendering time: 3.286511 seconds
rendering time: 4.971561 seconds

simplify:
rendering time: 1.123183 seconds
rendering time: 2.797506 seconds
rendering time: 3.292479 seconds
rendering time: 4.914665 seconds

no optim:
rendering time: 1.154879 seconds
rendering time: 2.925999 seconds
rendering time: 3.341494 seconds
rendering time: 5.098658 seconds

The optimization is almost useless, though we can keep Simplify which is always better.

Phyks:
For the record, difference between `ST_SnapToGrid` and `ST_Simplify` is that the latter can generate invalid geometries while the former should not. Then, usually `ST_SnapToGrid` should generally be preferred since it does not generate invalid geometries (and invalid geometries can be discarded by Mapnik and never rendered).

This is however true for closed ways (areas). I don't think any such issues could appear with lines (such as bicycle routes) so this is safe for a merge in my opinion.

Issue #65, opened 2019-02-20
Render conditionnal access to pedestrian or cycleway


Florimondable:
In Paris some streets are closed to trafic and declared pedestrian for some hours (sunday from 10 to 18 for instance) they are called "Paris Respire" area.

Could be good to know but hard to render I guess.

Florimondable:
In the same spirit, there are some cyleway which are closed some moments (the night for parks) or some mornings (market day).

See this old map for instance http://montreuil.fubicy.org/parivelo_mdb.html

May be a highlight in yellow like we do for cycle routes saying "warning, check the opening hours of the way"

Florimondable:
Real case for the mountain roads 
https://lists.openstreetmap.org/pipermail/talk-fr/2019-November/095128.html

the key is `[bicycle:]conditionnal=`
https://wiki.openstreetmap.org/wiki/Conditional_restrictions

Florimondable:
In very high zoom (like z20) we could draw in text along the road the `conditionnal `value.
For simple value `conditional `is human readable.

Florimondable:
I propose:
- draw thin *orange/red* line in around z17 in the middle of the road to warn that the road have bicycle access condition
- draw thin *blue* line in around z17 in the middle of the road to warn that the road have motor_vehicle access condition so the road is bicycle friendly some times.
- draw the conditional text along the road in high zoom (z19-20)

PR #311, opened 2020-02-18
Add conditional text along the way for bicycle|vehicle|access|motor_vehicle:conditional

Florimondable:
from z20
see #65 

results:
Capture du 2020-02-18 23-27-48
Capture du 2020-02-18 23-28-24
Capture du 2020-02-18 23-29-40
Capture du 2020-02-18 23-30-25

### DIFF
--- a/labels.mss
+++ b/labels.mss
@@ -364,6 +364,36 @@
   }
 }
 
+
+#roads-text-conditional[zoom >= 20] {
+  text-name: ""; /* Useless since it is voreloaded after, but it rise a warning. */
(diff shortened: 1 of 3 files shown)


Florimondable:
The result is ok, I think for the moment an axial line would be too much.
choose bicycle|vehicle|access|motor_vehicle:conditional because they are the most important for cycling (and the most used) https://taginfo.openstreetmap.org/search?q=%3Aconditional

Florimondable:
Color more red
Capture du 2020-02-19 21-49-20

PR #330, opened 2020-03-20
Railway at low zooms

### LINKED ISSUE #317: Railway in low zoom
In car oriented map we have the motorway in low zooms https://www.openstreetmap.org/#map=7/46.834/1.307
What are the motorway for cyclists ? Railways ! Railways help cyclist crossing a country quickly.
I use some time railway map to find how I can jump 100-500km quickly.
So my idea is to render railway in low zooms (may be =z10)
https://www.openstreetmap.org/#map=7/46.510/1.472&layers=T

Florimondable:
fix #317 

Capture du 2020-03-20 18-55-47
Capture du 2020-03-20 18-55-17
Capture du 2020-03-20 18-54-58
Capture du 2020-03-20 18-56-29

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -728,7 +728,13 @@ Layer:
       (
         SELECT
           ST_Snaptogrid(way, !pixel_width! / 4) AS way,
-          highway AS type,
+          COALESCE(
+            highway,
+            CASE
+              WHEN railway IN ('rail') THEN 'railway'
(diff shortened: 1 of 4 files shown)


Florimondable:
for the record I check a full country, it looks good :
Capture du 2020-03-21 16-23-25
Capture du 2020-03-21 16-29-15

PR #332, opened 2020-03-28
Render bridleway+bicycle=designated as a shared path.

### LINKED ISSUE #215: Footway + bicycle=yes is path
User issue:
https://www.reddit.com/r/openstreetmap/comments/d8mv1y/cyclosm_worldwide_release_at_sotm2019/f1wnomx/
«See for example https://www.cyclosm.org/#map=17/47.39015/8.40058/cyclosm along Bernstrasse there is an essentially perfect cycleway (https://www.mapillary.com/app/?lat=47.39439698&lng=8.40583019&z=17&pKey=yMkm9H1k-FXRo_XexcyMrQ&focus=photo), on the other hand the forest tracks nearby are just barely usable for cycling.»

Analysis :
https://www.openstreetmap.org/way/104156324#map=17/47.38849/8.39765
The tagging is may-be wrong here, it should be more a path.
But here lies the ticket if your read the footway page https://wiki.openstreetmap.org/wiki/Tag:highway%3Dfootway 
«Where a pathway is designated for pedestrians but is also allowed for bicycles you can use highway=footway and bicycle=yes.»
So footway with bicycle access tags should be rendered as we do for path.

Florimondable:
And refactor footway/bridleway + bicycle=designated|yes as path in sql.

fix #215

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -446,7 +446,10 @@ Layer:
           way,
           COALESCE(
             CASE
-              WHEN highway != 'bus_guideway' THEN highway
+              WHEN highway='raceway' THEN 'track'  -- render raceways as tracks
+              WHEN highway='footway' AND (bicycle='yes' OR bicycle='designated') THEN 'path'
+              WHEN highway='bridleway' AND (bicycle='yes' OR bicycle='designated') THEN 'path'
+              WHEN highway!='bus_guideway' THEN highway
               ELSE NULL
             END,
             CASE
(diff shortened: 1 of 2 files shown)


Florimondable:
I check the perf of the refactoring, the xml lost 0.6 Mo on 9Mo, same time in SQL request and may be a little faster for rendering test.

andrewharvey:
Thank you! This was the biggest issue I had with CycleOSM and now it looks fixed :heart:

PR #333, opened 2020-03-28
Lighten footway color in tunnel

Florimondable:
So we this can hide useless footways for cyclist (like subway tunnels).
I wonder if we could also use tags indoor and level for that

Screenshot_20200328_181331

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -460,7 +460,7 @@ Layer:
           0 AS bridge,
           access,
           layer,
-          1 as tunnel,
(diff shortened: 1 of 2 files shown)


Florimondable:
Use indoor tag to render way as tunnel, this should solve #334.

So with this train station, subway station, mall has there footway and steps faded, and don't interfere in the reading of the map for a cyclist.

Screenshot_20200329_163741
Screenshot_20200329_162900

PR #337, opened 2020-04-01
Render inner tube vending machine. Fix #315.

### LINKED ISSUE #315: Render inner tube vending machine
see https://wiki.openstreetmap.org/wiki/Tag:vending%3Dbicycle_tube

Phyks:
Shown from Z19 (argument behind this being that it is for a bicycle shop what an ATM is for a bank).

Icon still need some rework to make it lighter.


Florimondable:
I reworked the icon, see my pull request.

For the choice of the zoom level, I understand but there is vending machine alone without shop next to it like here 
https://www.openstreetmap.org/node/3048631403
https://www.openstreetmap.org/node/5562817463

So I think it'd valuable to render them from z16 or z15 like bicycle pump or a repair station but with a lower priority than a bicycle shop. (like we already do with those amenities)

PR #340, opened 2020-04-06
Render mtb scales on path

### LINKED ISSUE #336: Display MTB trail ratings
Would be nice to add some MTB trail information. I think the mtb:scale and mtb:scale:imba tags could be processed to add the appropriate coloring to the mtb trails
mtb:scale:imba=0 White (novice)
mtb:scale:imba=1 Green (easy)
mtb:scale:imba=2 Blue (moderate)
mtb:scale:imba=3 Black (difficult)
mtb:scale:imba=4 Red (expert)

Also could be considered is to add the some cooling to the route relations with the tag route=mtb as is done on the openstreetmap.org cycling map - they use orange.

Florimondable:
mtb:scale is rendered with dashes with STS colors (with simple, double or triple dashes).
mtb:scale:imba is rendered with dots with IMBA colors (double diamond is double dot).
If both is present mtb:scale is used.
Rendered on path (foot/bridle/cycle/path), track, service roads, from zoom 15.
fix #336 

STS :
Screenshot_20200406_160059
Screenshot_20200406_160035

IMBA :
Screenshot_20200406_155730
Screenshot_20200406_155600

both (rare case) :
Screenshot_20200406_160653

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fix a bug with some protected areas labels not rendered. See #329.
 * Render `highway=bridleway` + `bicycle=designated` as a shared path. See #215.
 * Render indoor paths in faded color. See #334.
+* Render `mtb:scale` and `mtb:scale:imba`. See #336.
 
 
 ## v0.3.4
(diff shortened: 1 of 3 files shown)


Phyks:
Can you update the legend as well please? =)

PR #350, opened 2020-04-23
Unknown quality surface / New colors / Fix surface over cycleway

### LINKED ISSUE #136: Dash road surface is rendered over cycle track
See there rue Tilsit https://cyclosm.github.io/cyclosm-cartocss-style/example.html#18/48.87447/2.29676

### LINKED ISSUE #324: Rework blue/cyan/green colors ?
It seems some people (how much ?) have difficulties to distinguish the different blue/cyan/green colors we use.

We don't perceive color the same way.
If I can trust this paper http://www.scilogs.fr/questions-de-couleurs/pourquoi-la-perception-des-couleurs-nest-elle-pas-parfaitement-identique-pour-chacun-de-nous/

Old people perceive less blue colors, cones for blue colors vary widely between people.
May be we should change the default color from blue to green, orange ?

We should remember the main design ideas :
- close color for fast reading (bluish == cyclable)
- different colors for detailed rendering
So it's a matter of balance.

(link to #171 though less extreme)

Florimondable:
This should fix #345, fix #324, fix #136 

Add unknown quality surface for path/track : simple one pixel dash line (visible but not too much)

This move the "foot color" from green to brown in order to have wider range of colors fixing #324
- footway : light brown
-    cycleway : darker blue
-    shared designated path : a little lighter blue
-    path : little greener
-    speed colors : wider shading to green/yellow
-    lighten casing for pedestrian and living streets, pedestrian are less visible (not bikefriendly)
-    redder bigger (as footway) for bridleways
-    grey nobicycle roads/motorway
-    fix maxspeed/no_motor color on footway-bridleway-steps
-    add no_motor color since pedestrian color is no more hard enough
-    darken a little track color
-    remove not used colors
-    Draw motorway in grey instead of white in low zooms.
-    Pedestrian with nomotor use pedestrian color instead of nomotor color, because (many) pedestrian is more important information than no vehicles at all for a cyclist.
-    darker road casing, for better contrast

Fix #136 by using attachment (`myattachment/line-style:*`) attachements of the same data is render at the time, not after like a layer.

Use attachment to draw road surface with little opacity, so there color is melted.

Path/track surfaces are rendered now with dash lines with a very light background (no more sub color per type) this make the quality surface reading easier and its less overwhelmed the map. It allows too to render surface from z12 ! Remove also one level of dashline, the code is simplier I think at the end.

Fix skinny motorways link at some levels.

I didn't change the color in the file road-colors.yml and didn't use the script because it's very boring, and useless ? Should be cleaned.

Here some examples :
Screenshot_20200423_221000
Screenshot_20200423_220948
Screenshot_20200423_220928
Screenshot_20200423_215805

Pedestrian, low maxspeeds
Screenshot_20200423_224003

Surface on roads, cycleway over road surface dash line fix
Screenshot_20200423_223903


Florimondable:
Don't merge yet, it's 50% slower, and xml is also 50% bigger :/
If you have ideas how to improve that...

Florimondable:
Last commit change the condition `[tunnel=1]` by the layer `#tunnel` the gain is 52s to 30 xml generation and 5Mo. Rendering is way faster too we are still about 10% slower in tile generation compare to previous version 0.3.5.
Compare to february version we have 1,5Mo 10s more xml generation, rendering is also slower by 35%.
Improvement is still needed.

project.mml
```diff
@@ -18,7 +18,7 @@ srs: "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0
 
 # Various parts to be included later on
 _parts:
-  # Extents are used for tilemill, and don't actually make it to the generated XML
+  # Extents are used for tilemill, and dont actually make it to the generated XML
```
Phyks, line 21: Typo, should be "don't" in English.

project.mml
```diff
@@ -48,8 +48,8 @@ Stylesheet:
 - ferry-routes.mss
 - power.mss
 - admin.mss
-- gpx.mss
-- legend.mss
+#- gpx.mss
+#- legend.mss
```
Phyks, line 52: Typo, these can be safely restored I guess.

roads.mss
```diff
@@ -1,25 +1,25 @@
-/* ==================================================================
-                    ROAD & RAIL LINES
-/* ================================================================== */
+// ==================================================================
```
Phyks, line 1: Is it just for beauty or does `//` vs `/*` have a broader impact?

roads.mss
```diff
@@ -1394,152 +1397,310 @@ come in as well.
...
+  [type='track'],
+  [type='bridleway'],
+  [type='footway'],
+  [type='path'],
+  [type='cycleway'] {
+    [surface_type='unknown'][zoom >= 13],
+    [surface_type='cyclocross'][zoom >= 12],
+    [surface_type='mtb'][zoom >= 12] {
+      background/line-join: round;
+      background/line-opacity: 0.4;
```
Phyks, line 1413: I'm missing something here, why is there an opacity for the background?

Phyks:
10% slower is fine. I have some questions but I think we can merge and work on optimizing afterwards.

project.mml
Florimondable: yeah but that ' break my color syntax :/

project.mml
Florimondable: but they are useless for the production, no ?

roads.mss
Florimondable: /* is much more slower when you test optimisation code by commenting big section of the file :D

roads.mss
Florimondable: Good question, with opacity the background can melt with the background map (landuse)

Florimondable: check tracks between forest and land https://user-images.githubusercontent.com/47089717/80146366-0b640a80-85b2-11ea-9f86-7d695fba2355.png

Florimondable:
> 10% slower is fine. I have some questions but I think we can merge and work on optimizing afterwards.

yes

project.mml
Phyks: In principle, this should not change much since they have `status: off`, isn't it?

roads.mss
Phyks: Wow

project.mml
Florimondable: ok

Phyks:
Merged manually.

PJDB22:
For a protanopia, there is still a problem for low speed limits. The color used for 30kph is to close to the white (residential/service). 20kph and under are ok for me.
In France, this speed limit as great consequence as it change the default meaning of one way sign to double way for cyclists.

Phyks:
Hi,

Thanks for the feedback, this is extremely useful to us! The colors currently used are defined in https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/road-colors.mss#L48-L49. This is:
- `#bbffff` for 30 kph ;
- `#bbffba` for 20kph.

If you are familiar with imaging, could you try, starting from this `#bbffff` value, to find a nearby color which would be clearly distinguishible for people with protanopia? We can also try to find something on our side and will likely ask you for feedback.

Thanks!

PJDB22:
After a few tests, I think #88ffff would be enougth to distinguish from white.
Thanks for your help!

Phyks:
Thanks for the feedback! `#88ffff` is very bright, so it might be too prominent for such a feature as 30kph speed.

I tried moving around in the color palette and I came across `#aff1f1` which seems to give a good result with some simulators. Here is a screenshot, may I ask a feedback on this?



Location is https://www.cyclosm.org/#map=18/48.86332/2.35234/cyclosm with the three maxspeed colors in the view.

Thanks!

PJDB22:
Thank you for your time, but the color are still to close..
For example on this place https://www.cyclosm.org/#map=18/48.51565/-2.77414/cyclosm
Clipboard01
The streets "Pinot Duclos" (speed limit 30) and "Tristant Corbière" (speed limit 50) look the same for me.

Florimondable:
Hi, the modification is not yet into production (on the website cyclosm.org).
It should be in this week :)

PJDB22:
It's very usable. 
Thank you!

PR #352, opened 2020-04-26
Lite CyclOSM variant

Phyks:
@Florimondable This should more or less produce the same output as #351, but with a global switch.

I'm using

```
kosmtik serve ~/Code/cyclosm-cartocss-style/project.mml --localconfig localconfig.js
```
for the standard render and
```
kosmtik serve ~/Code/cyclosm-cartocss-style/project.mml --localconfig localconfig-lite.js
```
for the lite variant.

Still some testing to be done.


Florimondable:
I don't realy like to maintain two styles... :/

PR #365, opened 2020-05-16
Add support for running the style with Docker(-compose)

hiddewie:
Taken from https://github.com/gravitystorm/openstreetmap-carto/blob/master/DOCKER.md

This guide makes setting up the database, mapnik, data and kosmtik much simpler, and without having to import data into the database running on the system.

See https://github.com/cyclosm/cyclosm-cartocss-style/blob/64461f470911da807bce852d5a8e106efabadeef/docs/DOCKER.md for textual documentation on getting started.

### DIFF
--- a/Dockerfile.db
+++ b/Dockerfile.db
@@ -1,0 +1,3 @@
+FROM mdillon/postgis:10
+
+ADD ./scripts/tune-postgis.sh /docker-entrypoint-initdb.d
--- a/Dockerfile.db
+++ b/Dockerfile.db
@@ -1,3 +1,3 @@
-FROM mdillon/postgis:10
+FROM mdillon/postgis
 
 ADD ./scripts/tune-postgis.sh /docker-entrypoint-initdb.d
--- a/docs/DOCKER.md
+++ b/docs/DOCKER.md
@@ -1,0 +1,79 @@
+# Running OpenStreetMap Carto with Docker
+
+[Docker](https://docker.com) is a virtualized environment running a [_Docker demon_](https://docs.docker.com/engine/docker-overview), in which you can run software without altering your host system permanently. The software components run in _containers_ that are easy to setup and tear down individually. The Docker demon can use operating-system-level virtualization (Linux, Windows) or a virtual machine (macOS, Windows).
+
+This allows to set up a development environment for OpenStreetMap Carto easily. Specifically, this environment consists of a
+PostgreSQL database to store the OpenStreetMap data and [Kosmtik](https://github.com/kosmtik/kosmtik) for previewing the style.
+
+## Prerequisites
+
+Docker is available for Linux, macOS and Windows. [Install](https://www.docker.com/get-docker) the software packaged for your host system in order
+to be able to run Docker containers. You also need Docker Compose, which should be available once you installed
+Docker itself. Otherwise you need to [install Docker Compose manually](https://docs.docker.com/compose/install/).
+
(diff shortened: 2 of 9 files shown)


Dockerfile.db
```diff
@@ -0,0 +1,3 @@
+FROM mdillon/postgis:10
```
hiddewie, line 1: In the install guide there is Postgres 9.6. Although locally I am running on Postgres 10.

docs/DOCKER.md
```diff
@@ -0,0 +1,83 @@
...
+* Ctrl+C to stop the style preview application
+* `docker-compose stop db` to stop the database container
+
+## Repositories
+
+Instructions above will clone main CyclOSM repository. To test your own changes you should [fork](https://help.github.com/articles/fork-a-repo/) cyclosm/cyclosm-cartocss-style repository and [clone your fork](https://help.github.com/articles/cloning-a-repository/).
+
+This CyclOSM repository needs to be a directory that is shared between your host system and the Docker virtual machine. Home directories are shared by default; if your repository is in another place you need to add this to the Docker sharing list (e.g. macOS: Docker Preferences > File Sharing; Windows: Docker Settings > Shared Drives).
+
+## Importing data
```
hiddewie, line 36: This might need some extra info on downloading shapefiles & elevation, and generating contours and hillshading.

scripts/docker-startup.sh
```diff
@@ -0,0 +1,70 @@
...
+  --cache $OSM2PGSQL_CACHE \
+  --number-processes $OSM2PGSQL_NUMPROC \
+  --hstore \
+  --database gis \
+  --slim \
+  -c \
+  -G \
+  --drop \
+  $OSM2PGSQL_DATAFILE
+  ;;
```
hiddewie, line 53: What about downloading height information and generating hillshade and contours?

Dockerfile.db
Phyks: Indeed, there should not be anything preventing CyclOSM to be run with PostgreSQL > 9.6. The INSTALL.md guide might night an update (maybe use the generic `postgresql` package without mentionning versions).

scripts/docker-startup.sh
Phyks: For contours, the easiest way to go is probably to use  `phyghtmap` which handles both downloading height information and generating contours, for a specific area. Note however that contours are not yet in production, we are in the process of importing them at the moment.

Provided you can download height information, generating hillshades should be a matter of gdal commands : https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/docs/INSTALL.md#create-hillshades.

Phyks:
Thanks for this PR!

There was a fork of a Docker tile server featured in OSM weekly some time ago, which includes CyclOSM as well, see https://github.com/mhajder/openstreetmap-tile-server-cyclosm (mentionned in https://github.com/cyclosm/cyclosm-cartocss-style#related-projects). What are the main differences? (might be worth emphasizing it in the README.me or DOCKER.md file).

docs/INSTALL.md
```diff
@@ -1,8 +1,7 @@
 Getting started
 ===============
 
-This document describes how to manually configure your system for running
-CyclOSM.
+This document describes how to manually configure your system for running CyclOSM. If you prefer quick, platform independent setup for a development environment, without the need to install and configure tools by hand, follow a Docker installation guide in [DOCKER.md](https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/docs/DOCKER.md).
```
Phyks, line 4: I would mention that preferred method is running CyclOSM manually. Docker alternative is provided as an alternative.

(reason being @Florimondable and I are mainly using the manual approach, so Docker approach is more community supported, although we'll do our best to update everything as needed)

hiddewie:
Thanks for the comments. I will improve this documentation.

I ran CyclOSM with the overv docker images (what the mhayder repository is cloned from) before, but the importing process was a less reliable than the docker compose file in the Carto style repository. Both work, and both produce roughly the same results.

A difference I like for this setup is the default configuration with a `db`, `import` and `kosmtik` image/container, which can be used with `docker-compose` without the need to remember environment variables and settings to start the import/tile-server Docker containers manually.

The Overv docker image also contains no kosmtik, and this has to be run by hand against the (Docker) tile server (I used `joxit/kosmtik` before).

Phyks:
Thanks!

Issue #366, opened 2020-05-17
Add maxzoom for layer state-names ?


Florimondable:
Max zoom state-names is drawn depending of the size of the state.
State is for adminborder level 4 https://wiki.openstreetmap.org/wiki/Tag:boundary=administrative (region in France)
Though I didn't found example where they are drawn >z12, so I propose to set maxzoom at 13

Here sql result for z16 :
```
./test_sql.sh ../../project.mml --zoom 16 --db osm ....
+-----------+----------------+-----------------------------+-----------------+---------------------------+
| Time (ms) | #lines fetched |            Layer            | #fields fetched | Size (text format, in kB) |
+-----------+----------------+-----------------------------+-----------------+---------------------------+
|    3486   |     516278     |          roads_high         |        20       |           150688          |
|    1164   |     13432      |        amenities-poly       |        31       |            9878           |
|    1004   |     100487     |       amenities-points      |        31       |           10860           |
|    444    |      1931      |      highway_area_label     |        5        |            180            |
|    431    |      4874      |         landuse_gen1        |        4        |           19519           |
|    408    |     33865      |   roads-text-ref-low-zoom   |        6        |            8696           |
|    346    |     10092      |          area_label         |        5        |           10900           |
|    332    |      1380      |       landuse-overlay       |        5        |            7096           |
|    296    |      122       |          water_gen1         |        4        |            1232           |
|    292    |       15       |       military-overlay      |        5        |             19            |
|    291    |       36       |       protected-areas       |        6        |            765            |
|    287    |       1        |         state-names         |        5        |            511            |
|    284    |       0        |        country-names        |        0        |             0             |
|    260    |     12601      |            bridge           |        19       |            1994           |
|    221    |     19590      |            tunnel           |        19       |            4998           |
|    139    |      7615      |       placenames-small      |        7        |            661            |
|    118    |      1564      |        admin-mid-zoom       |        3        |           17190           |
|    106    |       1        |        capital-names        |        5        |             0             |
|    105    |      1433      | bicycle_routes_bicycle_gen3 |        5        |            2440           |
|    102    |      252       |      placenames-medium      |        6        |             18            |
|     86    |       1        |   bicycle_routes_mtb_gen3   |        5        |             2             |
|     85    |      1339      |    bicycle_routes_labels    |        7        |            2277           |
|     83    |       15       |         ferry-routes        |        2        |             9             |
|     74    |      1435      |         waterway_low        |        3        |            570            |
[...]

PR #368, opened 2020-05-18
Labels zoom

### LINKED ISSUE #359: Render bike station and low small/low priority POI names from z18 instead of z17
This information is not important 99,9% of the time so in order to make the map clearer we can move it one level higher.

Florimondable:
(branch from optim-roads.mss)

Should fix #359 
Screenshot_20200518_204609
Screenshot_20200518_204539

### DIFF
--- a/amenities.mss
+++ b/amenities.mss
@@ -544,19 +544,6 @@
     }
   }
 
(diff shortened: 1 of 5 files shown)


Phyks:
Fixes #366

PR #408, opened 2020-07-11
docker import to "osm" database

quasart:
Hello,
I managed to run cyclosm docker but I had to change import script : osm data is imported to 'gis' database, but kosmtik is expecting 'osm' database


Phyks:
Hi,

Thanks for the fix!

Issue #400, opened 2020-06-21
cycleway surface quality rendering is rounded


Florimondable:
![Screenshot_20200621_165600](https://user-images.githubusercontent.com/47089717/85227854-2dc2aa80-b3e0-11ea-9ec4-1b86c84eba4c.png)

Issue #377, opened 2020-05-24
Render bench


Florimondable:
We render rest area or picnic table but not bench itself.

PR #411, opened 2020-07-18
Building contours and z14 track width

quasart:
Hello !

Now I can run cyclosm renderer, I tried to implement several minor changes.

Two opened issues :
* Render bench #377. Implemented exactly like defaut OSM map from zoom 19
* cycleway surface quality rendering is rounded #400

I pushed other cosmetic changes. Tell me If you don't want some of them, I can remove corresponding commits :

* Added very thin contour around building polygons in zooms 16/17 : contour had been removed for simplification but I found that even tiny contour was important  when land is gray

* Made cemetary pattern more dense, like OSM-fr style, because current pattern looks like POI icons.

* z14 track width was increased. Because it's the only zoom level where tracks and paths had the same width.

I hope it helps.

cosmetic - cemetary pattern + buidding contour
cosmetic - cycleway surface
cosmetic - z14 track width

### DIFF
--- a/amenities.mss
+++ b/amenities.mss
@@ -544,6 +544,16 @@
     }
   }
 
+  [feature = 'amenity_bench'][zoom >= 19] {
+    marker-file: url('symbols/openstreetmap-carto/amenity/bench.svg');
+    marker-fill: @man-made-icon;
+    marker-placement: interior;
+    marker-clip: false;
+    [access != ''][access != 'permissive'][access != 'yes'] {
+      marker-opacity: 0.33;
+    }
+  }
+
   [feature = 'aeroway_helipad'][zoom >= 16] {
     marker-file: url('symbols/openstreetmap-carto/helipad.16.svg');
     marker-placement: interior;
--- a/base.mss
+++ b/base.mss
@@ -204,10 +204,17 @@
 /* ---- BUILDINGS ---- */
 #buildings[zoom>=16] {
   polygon-fill: @building;
+  
+  /* Thin contour of buildings in mid zooms important for gray background areas */
+  line-color: darken(@building,10);
+  line-width: 0.3;
+  [zoom=17] {
+    line-width: 0.5;
+  }
 
   /* Render perimeter of buildings in high zooms */
   [zoom>=18] {
-    line-color: #b6b2af;
+    line-color: darken(@building,20);
     line-width: 0.75;
   }
 }
--- a/base.mss
+++ b/base.mss
@@ -23,7 +23,8 @@
   [type='landuse_cemetery'] {
     polygon-fill: @cemetery;
     [zoom >= 13] {
-      polygon-pattern-file: url('symbols/openstreetmap-carto/grave_yard_generic.svg');
+      polygon-pattern-file: url('symbols/openstreetmap-carto/grave_yard_generic_many.svg');
+      polygon-pattern-opacity: 0.15;
     }
   }
   [type='amenity_college']       { polygon-fill: @school; }
--- a/project.mml
+++ b/project.mml
@@ -1632,7 +1632,7 @@ Layer:
           COALESCE( -- order is important here
             'aeroway_' || CASE WHEN aeroway IN ('helipad', 'aerodrome') THEN aeroway ELSE NULL END,
             'tourism_' || CASE WHEN tourism IN ('artwork', 'alpine_hut', 'camp_site', 'caravan_site', 'chalet', 'wilderness_hut', 'guest_house', 'apartment', 'hostel', 'hotel', 'motel', 'information', 'museum', 'viewpoint', 'picnic_site', 'gallery') THEN tourism ELSE NULL END,
-            'amenity_' || CASE WHEN amenity IN ('atm', 'bank', 'bar', 'bicycle_rental', 'bicycle_parking', 'bicycle_repair_station', 'biergarten', 'cafe', 'car_wash', 'compressed_air', 'drinking_water', 'fast_food', 'ferry_terminal', 'food_court', 'fountain', 'hospital', 'ice_cream', 'internet_cafe', 'pharmacy', 'place_of_worship', 'police', 'post_office', 'post_box', 'pub', 'public_bath', 'restaurant', 'shelter', 'shower', 'toilets', 'water_point', 'cinema', 'theatre', 'bureau_de_change', 'casino', 'library', 'motorcycle_parking', 'charging_station', 'vending_machine') THEN amenity ELSE NULL END,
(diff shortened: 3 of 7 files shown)


Phyks:
Hi,

Thanks a lot for all these updates! I'm ok for a full merge on my side, I'll just wait for feedback from @Florimondable.

Best,

Florimondable:
> Hello !
> 
> Now I can run cyclosm renderer, I tried to implement several minor changes.
> 
> Two opened issues :
> 
>     * Render bench #377. Implemented exactly like defaut OSM map from zoom 19
Ok, no screenshot ?

> 
>     * cycleway surface quality rendering is rounded #400
Ok.

> 
> 
> I pushed other cosmetic changes. Tell me If you don't want some of them, I can remove corresponding commits :
> 
>     * Added very thin contour around building polygons in zooms 16/17 : contour had been removed for simplification but I found that even tiny contour was important  when land is gray

Not ok, what issue does it solve ? I mean, building is not very important for a cycling map.

> 
>     * Made cemetary pattern more dense, like OSM-fr style, because current pattern looks like POI icons.

Ok, cemetary is water point so more density make them easier to see.

> 
>     * z14 track width was increased. Because it's the only zoom level where tracks and paths had the same width.

Not Ok, same for service or residential, I see not glitch by switching zooms around 14.

quasart:
**Bench screenshot**

Screen Shot 2020-07-19 at 15 32 13


**Building contour**

I agree building should not take too much focus, but if you cannot see building shapes it is an important issue for localization.
* Exemple in hospital : https://www.cyclosm.org/#map=17/47.48239/-0.55559/cyclosm 
* Exemple in industrial zone : https://www.cyclosm.org/#map=17/47.48503/-0.54568/cyclosm


**z14 track width**

I don't know what to say, I can see it... It's definitely minor, but I've noticed there is a broken consistency, so why not fixing it ?
* https://www.cyclosm.org/#map=13/47.3097/-0.4707/cyclosm => track wider than path
* https://www.cyclosm.org/#map=14/47.3109/-0.4722/cyclosm => track and path look the same width
* https://www.cyclosm.org/#map=15/47.3110/-0.4782/cyclosm  => track wider than path

quasart:
Maybe you prefer to reduce path width instead of increasing track width ?

Florimondable:
> Maybe you prefer to reduce path width instead of increasing track width ?

I prefer nothing, roads width is a difficult balancing task, it needs to be tested with a dynamic map

quasart:
Ok, anyway let me know if I should change this PR

Phyks:
Hi,

Sorry for late feedback, I've been way more busy than expected lately (unexpected stolen bike…). Remaining points of discussion are building contours and z14 track width.

I've been confronted with the building issue very recently. Location was https://www.cyclosm.org/#map=16/48.7622/1.9885/cyclosm where it is next to impossible to spot buildings on the map (especially with a bad quality screen). Tagging is very scarce in this area, so buildings help a lot. I would be in favor of improvement on this side (either the fix from this PR or an alternative idea).

No strong opinion for me about the Z14 track width. We could merge the PR skipping this commit (thanks for the very clean PR with nicely separated commits!) to get the rest of the features very quickly.

Best,

Florimondable:
I'm ok with you @Phyks

Phyks:
Sorry for the delay on this one and thanks a lot @quasart.

I've just cherry-picked all commits except the two with remaining comments (track width and building contours). We definitely need something (especially for building contours). I don't have a good enough internet connection to make more tests at the moment, but will try to come back ASAP on these.

Florimondable:
If I'm not mistaken we can close this PR, modification has been done in other PRs ?

Phyks:
There are still two items which should be worked on:

* The homogeneization of the track/path width at z14 (do we want it?),
* Making buildings more visible.

For buildings, we definitely something IMO, see https://www.cyclosm.org/#map=16/48.7622/1.9885/cyclosm for instance where it is really hard to distinguish buildings from the background.

Phyks:
Coming back on this one, I think the fix of the track width at Z14 definitely makes sense.

See https://www.cyclosm.org/#map=14/45.7978/2.9871/cyclosm for instance and comparison with the screenshot with the fix applied:




South of "Puy de Pariou", it really helps making a distinction between the `highway=track` and the small `highway=path` portion. This definitely makes sense since one might expect the `highway=track` to be more bikable since a vehicle is expected to be able to pass!

Phyks:
Regarding the building contours commit, it is definitely a good fix for #565.

See https://www.cyclosm.org/#map=17/52.63799/13.28903/cyclosm




This is better but not enough for my example above. https://www.cyclosm.org/#map=16/48.7622/1.9885/cyclosm



The distinction looks too subtle to me to be readable in multiple conditions (in particular a bad smartphone screen / on a sunny day). 


It definitely does not make the buildings too heavy in dense urban areas, which is an excellent thing! https://www.cyclosm.org/#map=16/48.8156/2.3184/cyclosm

Phyks:
So, in short, I would:

* merge the fix for track width at Z14
* like some discussion on the buildings fix, maybe with a slight tweak of the outer line or of the industrial landuse color, before a merge.

rkflx:
> like some discussion on the buildings fix, maybe with a slight tweak of the outer line or of the industrial landuse color, before a merge.

Are colors chosen semi-randomly here, or is there an overall methodology that was defined beforehand and which is applied consistently? Guidelines might be more work in the beginning, but save a lot of frustrating tweaking in the long run.

Specifically, it might be helpful to think in terms of the HSL color model instead of RGB. Each map object (e.g. buildings, land use, ways) can be classified along multiple dimensions like type (area, line, polygon), physical extent, density, z-stacking, importance and so on. For each dimension a hierarchy of objects could be defined. There are also external expectations (e.g. grass==green, water==blue). The challenge is now to map dimensions and expectations to H, S and L channels in such a way that contrast for each channel is meeting needs (i.e. distances in the hierarchies are as high or low as required).

IOW, make the map distinguishable and balanced in grayscale, then colorize with colors from a predefined palette based on functional aspects.

Coming back to the buildings issue, I hope you'll see what I mean by analyzing the screenshots from above (i.e. let's find out why the added outlines are helpful, but the contrast issue is still not fixed completely):
- Open two color pickers with support for HSL (KColorChooser is quite helpful here) next to each other and compare building and land use colors in terms of HSL.
- Convert the screenshots to grayscale and look for contrast issues.
- Think about which dimensions are relevant for land use and buildings, define hierarchies and choose relative distances in S and L (e.g. land use A vs. land use B, and land use A/B vs. buildings).
- Finally, pick H to convey meaning.
- Note that outlines can be used as an additional tool (in a sense they warp S and L locally, allowing smaller distances).

(So far this is just an idea I had. If this approach turns out to work great for fixing the buildings issue, it might be worth applying it elsewhere too.)

---

Additional/alternative tip: There are websites used to assess contrast and legibility of various color combinations. Try to achieve at least the same score for contrast between buildings/outline and outline/land use as found on the standard osm.org style (i.e. reverse engineer there and apply findings here). I would suspect the outline could still be a bit stronger.

Phyks:
z14 track width fix has been merged into master.

Phyks:
Same for the building contour outline. This is improving the current situation and we have a numbe rof issues related to this topic. It's probably not perfect / enough, but it will be part of the next version (not the latest v0.6 which I've just tagged), so we'll have plenty of time to improve it.

Closing this one as everything has been handled!

PR #414, opened 2020-07-19
Fixes to the nomotor road style

### LINKED ISSUE #406: Render `motorcar=no motorcycle=no` like `motor_vehicle=no`
`motor_vehicle=no` is rendered as a bright green. But here in Switzerland, there is a distinction between that (tagged when [this sign](https://wiki.openstreetmap.org/w/images/5/5d/Verbot_f%C3%BCr_Motorwagen%2C_Motorr%C3%A4der_und_Motorfahrr%C3%A4der.png) is present), and `motorcar=no motorcycle=no` (tagged when [this sign](https://wiki.openstreetmap.org/w/images/8/86/Verbot_f%C3%BCr_Motorwagen_und_Motorr%C3%A4der.png) is present), which on Cyclosm is coloured white, the same as a street open to cars and motorcycles.

For bicycles, the only differences between the two signs are that roads with the second sign are also shared with 45km/h electric and motorised bicycles (in OSM tags, `mofa=yes`), while the first sign requires them to turn off their motor. But the blue signs for cycleways (`bicycle=designated`) also imply access for 45km/h vehicles. So I think it makes sense to use the same (green) colouring for both variants.

brainwad:
Update tags->'motorcar' to just motorcar.

motorcar is defined as a top-level field in planet_osm_line, and so not actually available in tags. This was breaking the nomotor styling for roads with motorcar=no.

EDIT (@Phyks): Fix #406.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -475,12 +475,16 @@ Layer:
             ELSE NULL
           END AS maxspeed_kmh,
           CASE
-            WHEN COALESCE(motorcar, tags->'motor_vehicle', tags->'vehicle', access, 'yes') != 'no' THEN 'yes'
-            -- goods and hgv are not technically children of motorcar, but the motorcar restriction is often intended to cover cars and all larger vehicles
-            WHEN COALESCE(tags->'goods', motorcar, tags->'motor_vehicle', tags->'vehicle', access, 'yes') != 'no' THEN 'yes'
-            WHEN COALESCE(tags->'hgv', motorcar, tags->'motor_vehicle', tags->'vehicle', access, 'yes') != 'no' THEN 'yes'
-            WHEN COALESCE(tags->'motorcycle', tags->'motor_vehicle', tags->'vehicle', access, 'yes') != 'no' THEN 'yes'
+            WHEN COALESCE(motorcar, tags->'motor_vehicle', tags->'vehicle', access, 'yes') NOT IN ('no', 'private') THEN 'yes'
+            -- goods and hgv don't need COALESCE chains, because the next step would be motorcar, which is checked above 
+            WHEN tags->'goods' NOT IN ('no', 'private') THEN 'yes'
+            WHEN tags->'hgv' NOT IN ('no', 'private') THEN 'yes'
+            -- moped and mofa are not checked, since most countries that have separate access controls for them treat them as quasi-bicycles
+            WHEN COALESCE(tags->'motorcycle', tags->'motor_vehicle', tags->'vehicle', access, 'yes') NOT IN ('no', 'private') THEN 'yes'
             -- TODO: style psv-only roads slightly differently
+            -- bus only needs to have its COALESCE chain go up to psv, because the next step would be motorcar, which is checked above
+            WHEN COALESCE(tags->'bus', tags->'psv') NOT IN ('no', 'private') THEN 'psv'
+            WHEN tags->'taxi' NOT IN ('no', 'private') THEN 'psv'
             ELSE 'no'
           END AS motor_vehicle,
           CASE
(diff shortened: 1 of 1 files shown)


brainwad:
This will fix #406.

Phyks:
Hi,

Thanks for the typo fix about `motorcar`! I however do not agree on the remaining changes. A `highway=footway` + `bicycle=designated` can be really different from a `highway=cycleway` (see https://github.com/cyclosm/cyclosm-cartocss-style/pull/414/files#diff-2d6c6d06f63a6fab47286addedefc335R458)!

See for instance https://www.openstreetmap.org/way/30168421 which is a dock made pedestrian. It is made for both pedestrian and cyclists, but does not provide the same comfort as a dedicated cycleway.

Same stands for segregated, it is simply an indication of design, but usage is usually that pedestrian and cyclist mix a lot and nobody understands segregated spaces (at least in France).

brainwad:
Sorry, those other commits aren't meant to be part of the pull request... They were more experimental. Not entirely familiar with GitHub :)

brainwad:
That said, I think the change I have for treating cycleway and bicycle=designated equually would be desirable. My main motivation is to precisely distinguish "true" cycleways from those which are shared with pedestrians and where one has to give way, as you mention. A `highway=cycleway foot=designated` indicates the second case, not the first case, despite being tagged as `highway=cycleway`, and a `highway=path bicycle=designated` is a synonym for `highway=cycleway` according to the wiki, and if it doesn't have `foot=designated` or `horse=designated` then bicycles probably have right of way.

The other fix, for `segregated` is more of a nit: the colours for a `highway=path segregated=yes` should be the same as if it were mapped as 2 separate ways: one dedicated cycleway, and one dedicated foot/bridleway. It's not a big deal, but I feel green should be reserved for paths where bicycles are allowed, and they aren't allowed on the foot half of a segregated path. My experience is that people mostly follow the segregation except when overtaking (at least in Switzerland, the USA, the Netherlands - where the surface of segregated paths is even different).

But I will try to fight with Git to remove these two commits from the PR and leave just the nomotor changes :) For these: what is your opinion on how roads where buses (only) are permitted? My feeling is that these are essentially traffic-free and so should be coloured a bike-friendly colour. I could be convinced buses are still loud, heavy and dangerous and so bus-and-bike-only roads should be treated the same as unrestricted roads. But that seems to loose a lot of fidelity (I'd rather ride with just buses, than with all traffic).

project.mml
```diff
@@ -475,12 +475,12 @@ Layer:
...
           CASE
-            WHEN (
-              tags->'psv' != 'no' OR tags->'motorcar' != 'no' OR tags->'bus' != 'no'
-              OR tags->'motor_vehicle' != 'no'
-              OR tags->'vehicle' != 'no'
-              ) THEN 'yes'
-            WHEN tags->'psv' IS NULL AND tags->'motorcar' IS NULL AND tags->'bus' IS NULL AND tags->'motor_vehicle' IS NULL AND tags->'vehicle' IS NULL AND (access != 'no' OR access IS NULL) THEN 'yes'
+            WHEN COALESCE(motorcar, tags->'motor_vehicle', tags->'vehicle', access, 'yes') != 'no' THEN 'yes'
+            -- goods and hgv are not technically children of motorcar, but the motorcar restriction is often intended to cover cars and all larger vehicles
+            WHEN COALESCE(tags->'goods', motorcar, tags->'motor_vehicle', tags->'vehicle', access, 'yes') != 'no' THEN 'yes'
```
Phyks, line 480: Couldn't this be reduced to a simple check on `tags->'goods'? The remaining should have already been handled by the first `WHEN` case.

project.mml
```diff
@@ -475,12 +475,12 @@ Layer:
...
+            WHEN COALESCE(tags->'hgv', motorcar, tags->'motor_vehicle', tags->'vehicle', access, 'yes') != 'no' THEN 'yes'
+            WHEN COALESCE(tags->'motorcycle', tags->'motor_vehicle', tags->'vehicle', access, 'yes') != 'no' THEN 'yes'
+            -- TODO: style psv-only roads slightly differently
```
Phyks, line 483: @Florimondable what is your opinion on this? PSV-only roads are basically `highway=busway` (this tag does not exist). Maybe these could be rendered as no motor vehicle?

E.g.: https://www.openstreetmap.org/way/202871519#map=18/48.72352/2.24382

project.mml
brainwad: Yeah, it could, but I thought leaving the full expression would make it easier to understand the logic (and to copy-paste if someone comes in in the future).

Florimondable: Please use the simplest version, we should taking care of time performance

project.mml
Florimondable: PSV-only roads have motor vehicle on it, so no-motor color can't be applied ;)

Your example is a no bicycle road btw

brainwad: Even if nomotor isn't the correct style, it could be worth defining a separate style for shared-with-PSV roads, similar to how cycleway=share_busway gets a style similar to, but inferior to, cycleway=lane. I've left a TODO, while making the current behaviour that these roads are styled as if cars were allowed on them.

PR #415, opened 2020-07-26
Treat use_sidepath as no

brainwad:
bicycle=use_sidepath denotes that cycling on the main road is legally forbidden, and that cyclists *must* use the sidepath. So it makes sense to render the same as bicycle=no, since it's just a specialisation that hints routers to look for a separate cycleway.


Phyks:
Thanks!

PR #421, opened 2020-08-08
Fix default access restrictions of highway=pedestrian, close #393.

### LINKED ISSUE #393: bicycle=dismount vs bicycle=yes
Hello all, 

New here, so apologies if this has already been discussed elsewhere. On hw=pedestrian, the default bicycle access value in many countries is dismount. Adding bicycle=yes (meaning cycling at modest pace) does not change the rendering of the road. I don't think that's ideal. Besides the risk cycling in pedestrian areas poses to pedestrians, local police also has an eye on that and likes to dish out fines. I would be very happy if the map told me whether I had to get off the bike or not. 
(Obviously that's a country-specific issue. But that's the case for other things, too.)

Best, John

Phyks:
`highway=pedestrian` with no extra tag are now greyed:

https://www.cyclosm.org/#map=18/48.87808/2.38647/cyclosm

`highway=pedestrian` + `bicycle=no|dismount` have the same rendering as no `bicycle` tag:

https://www.cyclosm.org/#map=18/48.87782/2.37984/cyclosm

`highway=pedestrian` + `bicycle=yes|designated` have the previous rendering with color fill (and possible overloading with speed limits):

https://www.cyclosm.org/#map=19/48.86115/2.35059

Closes #393.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Render `surface=unpaved` as unknown surface + minor fixes for surfaces.
     See #367.
 * Fix correct default access restrictions for `highway=pedestrian`. See #393.
+* Consider a way with `bicycle=dismount` is equivalent tagging to
+    `bicycle=no`.
 
 
 ## v0.3.6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Render `public_transport=platform` + `shelter=yes` as shelter. See #378.
 * Render `surface=unpaved` as unknown surface + minor fixes for surfaces.
     See #367.
+* Fix correct default access restrictions for `highway=pedestrian`. See #393.
 
 
 ## v0.3.6
--- a/project.mml
+++ b/project.mml
@@ -529,7 +529,7 @@ Layer:
             ELSE NULL
           END AS cycleway_right_oneway,
           CASE
-            WHEN bicycle IN ('no', 'private', 'use_sidepath') THEN 'no'
+            WHEN bicycle IN ('no', 'dismount', 'private', 'use_sidepath') THEN 'no'
             WHEN bicycle IS NOT NULL THEN bicycle
             WHEN tags->'motorroad' IN ('yes') THEN 'no'
             WHEN highway IN ('motorway', 'motorway_link') THEN 'no'
@@ -686,7 +686,7 @@ Layer:
             ELSE NULL
           END AS cycleway_right_oneway,
           CASE
-            WHEN bicycle IN ('no', 'private', 'use_sidepath') THEN 'no'
+            WHEN bicycle IN ('no', 'dismount', 'private', 'use_sidepath') THEN 'no'
             WHEN bicycle IS NOT NULL THEN bicycle
             WHEN tags->'motorroad' IN ('yes') THEN 'no'
             WHEN highway IN ('motorway', 'motorway_link') THEN 'no'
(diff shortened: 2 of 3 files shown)


DerDings:
I like the idea of being able to find pedestrian areas where bicycles are allowed on the map.

But I dislike the idea of using a grey fill for other pedestrian areas. Grey is used mainly for streets which cannot be accessed by cyclists, such as motorways or private ways.
Contrary to this, pedestrian areas are in reality very similar to footways, just larger. In many countries even the traffic signs for footways and pedestrian areas are (nearly) the same.

So for more consistency I think the default fill color of pedestrian areas should be as brown as footways are.
similar feature => similar colour

Florimondable:
I'm against coloring pedestrian highway in grey (like bicycle=no) by default, since this depends on local laws.

Can I close the PR ?

DerDings:
> 
> 
> I'm against coloring pedestrian highway in grey (like bicycle=no) by default, since this depends on local laws.
> 
> Can I close the PR ?

You're right, [wiki says](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dpedestrian#Vehicle_access) 
> "Cycling may be allowed, depending on the country or local restrictions. This can be specified with bicycle=yes or bicycle=no."

But drawing the way prominently in green by default also does assume something which cannot clearly be determined, just the other way round.

I think prominently highlighting ways as cycleable which are not suitable for cyclists in reality is worse than just not showing that cycling may be allowed in another road.

When looking at the traffic, pedestrian roads are more like a very broad footway rather than like a road. For cyclists, they're usually safe. But they're not suitable for long-way journeys, as you have to go slow. So I'd ditch the bright green entirely, and use path fill colors instead.

That would be 
`@path-fill` green if `bicycle=yes`
`@standard-nobicycle` grey if `bicycle=no`
`@footway-fill` brown as default, if `bicycle` is not given

It still would look sufficiently different from regular paths and footways, as the line is broader and has a prominent grey outline.

DerDings:
> When looking at the traffic, pedestrian roads are more like a very broad footway rather than like a road. For cyclists, they're usually safe. But they're not suitable for long-way journeys, as you have to go slow. So I'd ditch the bright green entirely, and use path fill colors instead.
> 
> That would be
> `@path-fill` green if `bicycle=yes`
> `@standard-nobicycle` grey if `bicycle=no`
> `@footway-fill` brown as default, if `bicycle` is not given
> 
> It still would look sufficiently different from regular paths and footways, as the line is broader and has a prominent grey outline.

I didn't manage to set up the development toolchain yet, but if you could hand me SVGs oft the three places given in the [initial comment](https://github.com/cyclosm/cyclosm-cartocss-style/pull/421#issue-465040049), I may make mock-ups to see if this conveys the information in a plain way.

Phyks:
@DerDings Here are few SVG tiles from these areas: http://pub.phyks.me/tmp/cyclosm/.

Let me know if you need anything else. Best,

Florimondable:
> But drawing the way prominently in green by default also does assume something which cannot clearly be determined, just the other way round.
> 
> I think prominently highlighting ways as cycleable which are not suitable for cyclists in reality is worse than just not showing that cycling may be allowed in another road.


I don't understand, for me the way is not drawing "prominently in green" it's a green-yellowish faded. It's really different than the bright green of a no-motor-vehicle road.

> When looking at the traffic, pedestrian roads are more like a very broad footway rather than like a road. For cyclists, they're usually safe. But they're not suitable for long-way journeys, as you have to go slow. So I'd ditch the bright green entirely, and use path fill colors instead.
> 
> That would be
> `@path-fill` green if `bicycle=yes`
> `@standard-nobicycle` grey if `bicycle=no`
> `@footway-fill` brown as default, if `bicycle` is not given
> 
> It still would look sufficiently different from regular paths and footways, as the line is broader and has a prominent grey outline.

No a pedestrian highway is a road, it's a street most of the time, it's wide, you can find vehicles there.
A footway brown would emphaze the road much more.

DerDings:
So this is how it would look like if [@footway-fill](https://github.com/cyclosm/cyclosm-cartocss-style/blob/4350a4635cab7df5e46007052babe3390c2fa51b/road-colors.mss#L19) was the default for pedestrian roads:

1_neu2_neu

3_neu
4_neu

6_neu
5_neu

DerDings:
> 
> I don't understand, for me the way is not drawing "prominently in green" it's a green-yellowish faded. It's really different than the bright green of a no-motor-vehicle road.
> 

To me it does look highlighted. And at least more similar to no-motor-vehicle roads than to other pedestrian-designated ways.
But that's the problem with individual color perception.

Florimondable:
I’m sorry but the footway color is way too strong for this kind of minor road.

If we want to change the pedestrian road color, we should go for a more faded yellowish color may be.

PR #422, opened 2020-08-15
turning_circle contour width

quasart:
Hi ! Here is a new minor cosmetic fix

I've notices turning circles contour line is always 1px width, but roads casing width increases at high zoom.
I changes this to make circles and residential roads share the same contour width.

turning_circle


Phyks:
Thanks!

PR #423, opened 2020-08-15
render brownish footway area

### LINKED ISSUE #370: Footway area are not drawn
example here https://www.cyclosm.org/#map=20/48.87813/2.28049/cyclosm

should be rendered like pedestrian (or may be less green more brown ? since footway is less cyclable than pedestrian ?)

quasart:
Footway area is rendered in light brown color :

in park
in city

Should close #370

### DIFF
--- a/base.mss
+++ b/base.mss
@@ -127,6 +127,7 @@
   }
   [type='amenity_parking']       { polygon-fill: @parking; }
   [type='highway_pedestrian']    { polygon-fill: @pedestrian_area_fill; }
+  [type='highway_footway']       { polygon-fill: @footway_area_fill; }
   [type='landuse_religious']     { polygon-fill: @religious; }
   [type='leisure_pitch']         {
     polygon-fill: @sports;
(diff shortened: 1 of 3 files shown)


Phyks:
Thanks, this looks great! I like the brownish feel for such places, as they tend to be sand / compacted surface and it makes sense. We should however probably double check how this would look like in city, such as the example of the square in front of Montparnasse: https://www.cyclosm.org/#map=20/48.87813/2.28049/cyclosm (this is however a **very** difficult place for renders, even osm.org render is having troubles there). Another example is Montrouge (https://www.cyclosm.org/#map=17/48.81876/2.31623/cyclosm) where sidewalks are tagged with `highway=footway` areas.

I don't have a reliable internet connection to do these checks myself for the next two weeks. I'll have a deeper look at these PRs as soon as I recover such a connection.

Best,

Florimondable:
I'm still wondering if we should use pedestrian area color or an other one ?

Phyks:
With only roads and landuse layers:




With all the layers:





Render is really good in my opinion, and subtle enough when this is actually a sidewalk. Let's get this merged! Thanks!

Issue #401, opened 2020-07-02
path colors


quasart:
Hello Cyclosm !

I try to understand path rendering.

Reading the legend, I'm expecting :
* Blue is cycleway (track where cycling should be nice and obvious).
* Cyan is `bicycle=designated` but not fully a cycleway (This one is not that simple to understand). De facto it's rendering of path made for cycling but shared with pedestrian (not segregated).
* Green is general purpose path (open to bikes but not made for them).
* Brown is footway/bridleway (path exists, but you should not ride there).
* Dark blue is a track (something like a general purpose path, but large).

I have some questions, because we meet many tag combinaisons in OSM database :

1. `highway=cycleway foot=yes/designated` is rendered blue. I thought it's bad tagging to have pedestrian on a cycleway, but this is allowed [here](https://wiki.openstreetmap.org/wiki/Tag:bicycle=designated), and it's common to describe French "voie verte".
Don't you think it should be rendered like a `highway=path/footway bicycle=designated` ? Blue cycleway with a footway border when `segregated=yes`, Cyan otherwise.

2. About this footway border, it's rendered in green today. It should have changed to brown like footway did. Am I wrong ?

3. `highway=path bicycle=no` is rendered in gray, making them hardly visible in parks. But this tagging sounds like footway. Don't you think all public footways should be rendered the same way for consistency ?

4. `highway=track` is dark blue, very close to cycleway color. In my opinion, it should look more green, somehow like general purpose path. What do you think ?

5. `highway=track bicycle=designated` could be rendered in cyan like any designated path, no ?

Cheers

Florimondable:
1. no, a cycleway is the best infra for cyclist even if there is sometime pedestrian on it
2. where ?
3. path are not footway (even with bicycle=no)
4. for me the two blue are not close (but that depend of people), may be more like a path, the point with track is that they should not have much trafic sinc it's for agricultural or forestry, that why it's more blue.
5. no track can have motor vehicle not path, they should have blue arrows as we already do on other roads.

quasart:
2. An example there : https://www.cyclosm.org/#map=19/48.87389/2.23633/cyclosm

3. Slightly different in OSM datamodel documentation. But today `highway=path access=no` and `highway=path bicycle=no` are rendered the same way. From a cyclist point of view : you can dismount and walk in the second case, not the first. Rendering foot-only path like footway is an acceptable and understandable simplification IMHO

brainwad:
[...]

PR #427, opened 2020-08-30
Greener tracks

brainwad:
Shift the hue of the tracks from 214° (very close to the hue of mixed-cycle-fill, which is 217°) to 140° (matching the hue of nomotor-fill), while maintaining the lightness and saturation. For reference, the hue of path-fill is 170° and cycle-fill is 240°, so with this change there is a continuum from roads that are (mostly) closed to motor vehicles (greenest) through paths that are either allowed or designated for bikes, but not reserved, to dedicated cycleways (bluest).

base vs 
this change.

The screenshots show a variety of backgrounds. I think the tracks are still quite visible even over forests, but it does stand out less there than the blue does. On other backgrounds (grass, farmland, plain) they are about the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Render `public_transport=platform` + `shelter=yes` as shelter. See #378.
 * Render `surface=unpaved` as unknown surface + minor fixes for surfaces.
     See #367.
+* Render `highway=track` as dark green, instead of dark blue.
 
 
 ## v0.3.6
--- a/road-colors.mss
+++ b/road-colors.mss
@@ -48,7 +48,7 @@
 @speed20-fill: #bbffba;
 @speedWalk-fill: #ddffba;
 @nomotor-fill: #62ff96;
-@track-fill: #1b6634;
+@track-fill: #114021;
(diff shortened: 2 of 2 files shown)


brainwad:
This would partially address #401

Florimondable:
Hi, thanks for the work, I like the idea.
I choose a dark blue because :
- must be visible easily (I like gravel biking) in the country and in the wood at low zoom (like here https://www.cyclosm.org/#map=12/49.3189/4.3005/cyclosm_live), so you can trace a route easily for tenth of kilometers.
- I understand that since they are shared road they should be more green than blue

So, can you make a try with a darker green, colors is difficult because to choose, our color perception is not as easy as math ;)

brainwad:
I matched the contrast ratio of the old colour against the forest background (5.3:1), and it seems to be "as dark" to me now. The green-ness is less pronounced, of course, but it's still further away from the cycleway colour than before. Sorry I don't have France downloaded, but I tried to find a similar area of Switzerland :)

z12
z15

Phyks:
@Florimondable What's the status on this one? Is it a go for a merge? (I see a thumb up on latest screenshot)

Florimondable:
It was not a go for a merge. Nevermind, I may going to tweak the colors
.

Florimondable:
Merged !
Nice to see it on a live map, green tracks make them more understandable and look nice, thank you for your work.

PR #430, opened 2020-09-09
 Render lcn routes tagged on ways directly

brainwad:
LCN routes are often tagged with a simple `lcn=` tag on the ways, rather than with route relations, especially because they often aren't numbered/named. Other tiers of cycle routes are much-less-often tagged this way, see taginfo: [lcn](https://taginfo.openstreetmap.org/keys/lcn) [rcn](https://taginfo.openstreetmap.org/keys/rcn) [ncn](https://taginfo.openstreetmap.org/keys/ncn) [icn](https://taginfo.openstreetmap.org/keys/icn).

The most common value is `yes`, but the next most common is `proposed`, which should be rendered similarly to relation-based routes with `state=proposed`, presumably.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1126,7 +1126,7 @@ Layer:
           UNION ALL
           SELECT way, 'lcn' as type, tags->'lcn' AS state, 4 as priority
           FROM planet_osm_line
-          WHERE tags->'lcn' IS NOT NULL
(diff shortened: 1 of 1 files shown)


Phyks:
Hi! Sorry, this one slipped out of my focus. Looks good to me, merging! Thanks!

PR #432, opened 2020-09-10
Place cycleways and paths at top of z-order hierarchy

### LINKED ISSUE #431: Shared cycleway rendered under road tagged with railway=abandoned
The z-order for paths should be higher than roads, so they are not clipped at medium zoom. In this picture, the path continues as far as the entrance to the driveway loop next to the restaurant, but it's barely visible beyond the cemetary, due to being quite close to the road:
image

brainwad:
Promote cycleways / paths with bikes allowed to z_order 99 and other paths to 98. This ensures they always render on-top of roads at the same layer, even when those roads are also tagged with railway= and so get a +35 boost to their z_order (see https://github.com/openstreetmap/osm2pgsql/blob/a424919b7810922be5586d6e24cf25126b4557ad/src/tagtransform-c.cpp#L61 for this behaviour). 

Note that tram tracks are still drawn on-top of cycleways/paths, as are railroad tracks, even after this change. This fixes #431 .

z17
z14

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -578,8 +578,8 @@ Layer:
             ELSE 'unknown'
           END AS surface_type,
           CASE
-            WHEN highway='cycleway' OR (highway IN ('path', 'footway', 'pedestrian', 'bridleway') AND bicycle IN ('yes', 'designated')) THEN CASE WHEN layer~E'^\\d+$' THEN 100*layer::integer - 61 ELSE -61 END
-            WHEN highway IN ('path', 'footway', 'pedestrian', 'bridleway') THEN CASE WHEN layer~E'^\\d+$' THEN 100*layer::integer - 62 ELSE -62 END
+            WHEN highway='cycleway' OR (highway IN ('path', 'footway', 'pedestrian', 'bridleway') AND bicycle IN ('yes', 'designated')) THEN CASE WHEN layer~E'^\\d+$' THEN 100*layer::integer - 1 ELSE -1 END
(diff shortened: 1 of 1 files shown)


Phyks:
Looking at https://github.com/openstreetmap/osm2pgsql/blob/e028bd19e098382102809c77d4400049ddc9da69/src/tagtransform-c.cpp#L13-L33, it seems the maximum `z_order` would be 39 plus eventually 35 in case there is a `railway` tag. Hence, shouldn't 75 be enough to ensure cycleways are always on top?

> Note that tram tracks are still drawn on-top of cycleways/paths, as are railroad tracks, even after this change.

Are you sure? They should be drawn in the same `roads_high` layer so I guess there could be an issue here. Could you point me to typical railways / tram tracks you checked? Thanks

brainwad:
Yes, but 99 and 98 are more future-proof: on a cycle map, I think it's fair to assume that cycleways should be the top-rendered highway in any given layer.

Here are screenshots of a) a railway crossing a designated path; b) a tramway crossing a cycleway. Both before and after this change, the tracks always render on top. I agree it's not intuitive, as the z_order of the railways are only 35. I checked and they *are* in the same layer (roads_high).

image
image

Phyks:
> I agree it's not intuitive, as the z_order of the railways are only 35. I checked and they are in the same layer (roads_high).

Indeed, that is expected actually since railways have a z_order of 35 and cycleways used to have at least 38.

Good for me then, let's get this merge.

Issue #162, opened 2019-09-21
Very bright green parking lot


tordans:
I wonder why those two parking lots are so bright green?
- https://www.cyclosm.org/#map=17/52.48683/13.42457/cyclosm
- OSM: https://www.openstreetmap.org/way/60825550

Its looks more green that the "Pedestrian" design (https://www.cyclosm.org/legend.html)? But this does also not fit the tagging(?).

IMO those parking spaces should be neutral in color, they are nothing special form a cycle point of view.

Phyks:
This one has `maxspeed=5` which according to OSM wiki means everyone should drive at walking speed. Therefore, it is considered by CyclOSM as a pedestrian area and very bike friendly.

Florimondable:
A precisision : it's not considered as pedestrian area, but only as a very pedestrian and bike friendly (just based on maxspeed tag).

I can see one thing missing : there is no information on cyclosm that it is a parking lot (good to know to locate yourself). -> add a P icon ?

The tagging of the way don't say it is a _service=parking_aisle_, and we don't specialized the drawing with the tag service. That could be a good idea to render them smaller.
And we could deduce the bikability of a service highway with the _service_ tag, a _service=emergency_access_ should be very bikable for instance.

PR #437, opened 2020-10-04
Hidden and thiner minor service road

### LINKED ISSUE #372: Rendering of service subtypes with thinner lines
As it is done by the standard openstreetmap.org rendering engine, would be great that the highway=service with a service=(driveway, drive-through, parking_aisle) be rendered with a line that is much thinner than the default highway=service road. In particular driveway going to private houses and parking lot aisle wound be distinguishable and yet much less dominant in the map. The current rendering make no distinction and in many areas where the thinner subtypes are being used it is not very readable. # #

quasart:
fix #372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Render footway areas with a brownish tone. See #370.
 * Render linear `highway=ford`. See #424.
 * Making ferry routes that can carry bikes more blue.
+* Making minor service roads less visible. See #372.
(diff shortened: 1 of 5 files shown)


quasart:
New commit to add 'P' icon on parkings (#162)

Depends on zoom level : Only big parking areas have their icon displayed in mid zoom.

Screen Shot 2020-11-08 at 17 53 17
Screen Shot 2020-11-08 at 17 53 35
Screen Shot 2020-11-08 at 17 53 53

Florimondable:
I don’t think it’s usefull to have parking P icon at z14 and z15, even z16.
And I would try with a smaller icon.

PR #438, opened 2020-10-06
fix proposed route no more rendered with dashes

quasart:
Hello,
Great to see last updates released, thank you!

But I have noticed a new issue : routes with `state='proposed'` are no more dotted.
Here is a fix

Screen Shot 2020-10-07 at 00 09 25


Phyks:
Thanks for the fix!

PR #447, opened 2020-10-31
Fix lowzooms : remove motorway, cycle routes transition, roads at z11

Florimondable:
no more motorway :
Screenshot_20201031_201028

Thinner cycle routes, and secondary are in orange color (case color), railway can be well distinguish :
Screenshot_20201031_201004

Screenshot_20201031_200932

Roads at Z11 are now visible :
Screenshot_20201031_201450

Screenshot_20201031_200742

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -80,6 +80,7 @@ overlapping borders correctly.
       line-color: @admin-boundaries;
       line-join: bevel;
       line-width: 0.3;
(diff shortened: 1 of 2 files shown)


Florimondable:
Render admin_level=4 (states/regions) with dash from z4 to avoid mixed them up with railway :

Screenshot_20201101_215526

PR #450, opened 2020-11-13
Better segregated rendering

brainwad:
This PR does two things:

First, it changes the colour of the pedestrian side-way, rendered at z17+ when `segregated=yes`, away from green. Green would have made sense before footways were shifted from green to brown, but now it doesn't so much - if two separate lines are rendered, the non-bicycle one should be rendered as non-bicycle ways are normally: brown. The colour of a bridleway is used if the way supports horse-riders, otherwise the colour of a footway:

TODO pic with foot
TODO pic with horse

Secondly, and dependent of the first change, it adds support for rendering the side-way for segregated paths that don't have `bicycle=designated` (designated is only tagged in some countries when a path legally requires usage; in these countries it doesn't imply anything about the quality of the path for cyclists, and a `bicycle=yes segregated=yes` path could be equal or higher-standard). This would have been confusing before the colour change, because it would have just rendered a path-coloured side-way next to the path-coloured main way, which would have just looked like a fat path, not two side-by-side ways. Now with the brown colour it works:

TODO pic with path

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -756,23 +756,29 @@
 }
 
 
-#roads_high::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated!='no'],
-#tunnel::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated!='no'],
-#bridge::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated!='no'] {
+#roads_high::path_outline_right[zoom>=17][type='path'][can_bicycle!='no'][segregated!='no'],
+#tunnel::path_outline_right[zoom>=17][type='path'][can_bicycle!='no'][segregated!='no'],
+#bridge::path_outline_right[zoom>=17][type='path'][can_bicycle!='no'][segregated!='no'] {
   line-cap: butt;
   [segregated='foot'] { line-color: @footway-fill; }
   [segregated='horse'] { line-color: @bridleway-fill; }
 
   line-width: @rdz17_path;
-  line-offset: @rdz17_path/2 + @rdz17_cycle/2;
-  [oneway='no'][oneway_bicycle='no'] {
-    line-width: @rdz17_path; line-offset: @rdz17_path/2 + @rdz17_cycle*1.5/2;
+  line-offset: @rdz17_path;
+  [can_bicycle='designated'] {
+    line-offset: @rdz17_path/2 + @rdz17_cycle/2;
+    [oneway='no'][oneway_bicycle='no'] {
+      line-width: @rdz17_path; line-offset: @rdz17_path/2 + @rdz17_cycle*1.5/2;
+    }
   }
   [zoom>=18] {
     line-width: @rdz18_path;
-    line-offset: @rdz18_path/2 + @rdz18_cycle/2;
-    [oneway='no'][oneway_bicycle='no']{
-      line-width: @rdz18_path; line-offset: @rdz18_path/2 + @rdz18_cycle*1.5/2;
+    line-offset: @rdz18_path;
+    [can_bicycle='designated'] {
+      line-offset: @rdz18_path/2 + @rdz18_cycle/2;
+      [oneway='no'][oneway_bicycle='no']{
+        line-width: @rdz18_path; line-offset: @rdz18_path/2 + @rdz18_cycle*1.5/2;
+      }
     }
   }
 }
--- a/roads.mss
+++ b/roads.mss
@@ -756,11 +756,12 @@
 }
 
 
-#roads_high::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated='yes'],
-#tunnel::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated='yes'],
-#bridge::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated='yes'] {
+#roads_high::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated!='no'],
+#tunnel::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated!='no'],
+#bridge::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated!='no'] {
   line-cap: butt;
-  line-color: @path-fill;
+  [segregated='foot'] { line-color: @footway-fill; }
+  [segregated='horse'] { line-color: @bridleway-fill; }
 
   line-width: @rdz17_path;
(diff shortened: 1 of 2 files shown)


roads.mss
```diff
@@ -756,11 +756,12 @@
...
 
 
-#roads_high::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated='yes'],
-#tunnel::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated='yes'],
-#bridge::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated='yes'] {
+#roads_high::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated!='no'],
+#tunnel::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated!='no'],
+#bridge::path_outline_right[zoom>=17][type='path'][can_bicycle='designated'][segregated!='no'] {
   line-cap: butt;
-  line-color: @path-fill;
```
Florimondable, line 763: Should have a default color (footway).
Should always has a default value, otherwise this makes more rules, hence more time to generate the xml and to render.

brainwad:
After taking the screenshots, I can barely tell the difference between bridleway and footway shading. Also, while path's default `horse` is `yes`, in practice most paths tagged with `segregated` and `bicycle` do not ever see horses. Should I remove the horse support and just always use footway colour? Should we do that for all bridleways?

Florimondable:
Hm, I think we should keep thing simple :
If bicycle=designated AND seggregated=yes -> cycleway + path rendered next to it
else path rendering.
Other condition are not related to cycling.

Brown is the color for pedestrian, red brown for horseman.

brainwad:
But like I mentioned, `bicycle=yes segregated=yes` is a thing. `designated` in many countries means that the law requires you to use the path, not that it's of a certain quality. A path with `bicycle=yes` can be equal or better quality than a path with `bicycle=designated`. And since non-designated paths are green, without the change to brown for the side-path, it will be illegible.

brainwad:
Also: I don't really get why the map has both brown and red-brown for foot/horse... This is a cycle map, surely one shade of brown for both is fine?

Florimondable:
> Some examples of non-`designated` segregated paths that I have tagged recently, on Mapillary:
> https://www.mapillary.com/map/im/cAGs_AO9jDMCWBK5yWfAOw
> https://www.mapillary.com/map/im/EBqLADoy4cvkpDyNdIaKpQ
Those path are designated for cyclist, the bicycle logo on the ground.

Briddleway rendered differently than footway, because most of the time the briddleway I know are made of sand, which is horrible for bicycle, so even without knowing its surface its good to have a warning about it.

brainwad:
Not per the tagging guidelines for Switzerland, which only tag designated
when the blue signs are used. The second example is not required to be
used, unlike a true designated cycleway.

Florimondable:
Designated value doesn't mean the way is compulsory.
If so the value use_sidepath can be used https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Duse_sidepath.

brainwad:
See https://wiki.openstreetmap.org/wiki/Switzerland/Map_Features.
Switzerland isn't the only country that tags this way, see also Belgium (
https://wiki.openstreetmap.org/wiki/WikiProject_Belgium/Conventions/Cycleways),
Germany (
https://wiki.openstreetmap.org/wiki/DE:Bicycle/Radverkehrsanlagen_kartieren
).

Florimondable:
I can't find in the pages that designated should be use when it's compulsory.

brainwad:
BE: "All roads/paths which do not display the signs depicted above, but
where cycling is allowed are best tagged with highway
<https://wiki.openstreetmap.org/wiki/Key:highway>=* (use: unclassified,
path etc, depending upon the basic road type) and then bicycle
<https://wiki.openstreetmap.org/wiki/Key:bicycle>=yes
<https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Dyes> if explicitly
signed"

CH: highway=path, other tags can be added  way  All other ways[1].  Without
one of above sign

In combination with
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dpath#Usage_as_a_universal_tag
, which specifies that on paths, designated is used for specifying the
legal designation of a path.
https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Ddesignated supports this
view ("Typically bicycle=designated (or highway=cycleway) requires
special-purpose traffic signs[1]. ")

DE:

*Radfahrer frei*
Zusatzzeichen 1022-10 alleine an einem Weg.

highway <https://wiki.openstreetmap.org/wiki/DE:Key:highway>=*(wenn von
anderen Fahrzeugen benutzbar meist track
<https://wiki.openstreetmap.org/wiki/DE:Tag:highway%3Dtrack> oder service
<https://wiki.openstreetmap.org/wiki/DE:Tag:highway%3Dservice>, sonst path
<https://wiki.openstreetmap.org/wiki/DE:Tag:highway%3Dpath>)
foot <https://wiki.openstreetmap.org/wiki/DE:Key:foot>=yes
<https://wiki.openstreetmap.org/wiki/DE:Tag:foot%3Dyes>
bicycle <https://wiki.openstreetmap.org/wiki/DE:Key:bicycle>=yes
oneway <https://wiki.openstreetmap.org/wiki/DE:Key:oneway>=yes/no [3]
<https://wiki.openstreetmap.org/wiki/DE:Bicycle/Radverkehrsanlagen_kartieren#cite_note-onewayOptional-3>
traffic_sign <https://wiki.openstreetmap.org/wiki/DE:Key:traffic_sign>
=DE:1022-10

VladimirMorozov:
> 
> Some examples of non-`designated` segregated paths that I have tagged recently, on Mapillary:
> https://www.mapillary.com/map/im/cAGs_AO9jDMCWBK5yWfAOw
> https://www.mapillary.com/map/im/EBqLADoy4cvkpDyNdIaKpQ

Hello, just want to be sure. These yellow road markings don't hold the same meaning as the round blue signs, correct? In my country similar road markings would have the same meaning as signs, that's why I'm asking.

brainwad:
@VladimirMorozov without the blue signs they are just advisory - you _may_ use them, but don't _have_ to. With the signs, they are compulsory - you aren't allowed to ride on the road anymore, but must use the path. It also has some consequences for electric scooters/skateboards, fast electric bikes ("s-pedelecs") & light mopeds ("mofas"): the blue signs allow them to use their motor while without it they can't.

kaneap:
I believe that any path with segregated=yes should be rendered the same regardless of whether it is bicycle=yes or bicycle=designated (dark blue= exclusive cycle path, which is implied by segregated=yes, regardless of whether you must use it)

PR #453, opened 2020-11-14
Fix bug in motorway_link colour

### LINKED ISSUE #452: Motorway Link are red instead of grey (Bicycle are not allowed there)
There: https://www.cyclosm.org/#map=19/48.81742/2.41731/cyclosm_live

image

On this picture, the grey and red roads are forbidden to bicycles.

The `motorway_link` is (kinda) red (Probably "Primary" -"Highway infrastructure") but "Bicycle not allowed" is "grey" and according to https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions#Default  for `motorway_link`, `bicycle=no` by default.

image

image

**So by default, motorway_link should be grey are they are forbidden for bicycles.**

brainwad:
The condition for the motorway_link (and trunk_link) fill colour was the wrong way around, so that it was shown as cycleable unless bicycle=yes.

Fixes #452.

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -2181,7 +2181,7 @@
     line-join: round;
     line-color: @motorway-trunk-cycle-fill;
     #tunnel { line-color: lighten(@motorway-trunk-cycle-fill, 10%); }
-    [can_bicycle='yes'] {
+    [can_bicycle='no'] {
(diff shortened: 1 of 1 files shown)


Phyks:
Thanks!

Issue #454, opened 2020-11-15
Private toilets are not rendered as such (dimmed)


Phyks:
See https://www.cyclosm.org/#map=20/48.81878/2.32169/cyclosm.

![2020-11-15-160928](https://user-images.githubusercontent.com/3856586/99188588-f4a72000-275c-11eb-8ba6-7ded34ad2936.png)

Florimondable:
fixed by #455

PR #455, opened 2020-11-16
Fix drinking water, toilets, acess

### LINKED ISSUE #399: More visible drinking water icon ?
«J'ai une question concernant l'affichage des points d'eau et toilettes.
J'ai remarqué qu'ils apparaissent au niveau de zoom 16, comme dans la carte cycle "historique". Mais  des qu'on passe au zoom 17, quasiment toutes les autres icônes apparaissent et ils sont noyés en zone dense, par exemple un centre-ville. La simple différenciation de couleur bleu / gris ne suffit pas pour les distinguer facilement au milieu des autres informations.
Serait-il possible de trouver une autre couleur ou même une autre icône plus visible ?  La solution de la carte historique est un peu moche avec son icône carrée opaque, il faudrait trouver un intermédiaire.»

Florimondable:
* Rework drinking water icon
* Distinguish toilets with drinking water.
* Distinguish access for toilets and drinking water.

Fix #399 #454

Screenshot_20201116_142501
Screenshot_20201116_142414
Screenshot_20201116_142325

### DIFF
--- a/palette.mss
+++ b/palette.mss
@@ -173,7 +173,7 @@
 @gastronomy-text: darken(@gastronomy-icon, 5%);
 @shop-icon: #666;
 @shop-text: darken(@shop-icon, 5%);
-@amenity-water: #007ae2;
+@amenity-water: #3b029a;
 @amenity-common: #666;
 @memorials: @amenity-common;
 @man-made-icon: #666;
--- a/symbols/openstreetmap-carto/amenity/drinking_water.svg
+++ b/symbols/openstreetmap-carto/amenity/drinking_water.svg
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns="http://www.w3.org/2000/svg"
+   version="1.1"
+   width="14"
+   height="14"
+   viewBox="0 0 14 14">
+  <rect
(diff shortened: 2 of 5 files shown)


symbols/openstreetmap-carto/amenity/drinking_water.svg
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
Phyks, line 1: This file should be deleted, isn't it?

palette.mss
```diff
@@ -173,7 +173,7 @@
 @gastronomy-text: darken(@gastronomy-icon, 5%);
 @shop-icon: #666;
 @shop-text: darken(@shop-icon, 5%);
-@amenity-water: #007ae2;
+@amenity-water: #3b029a;
```
Phyks, line 176: Is it really better to have this dark purple color instead of grey as `@health-color`?

Florimondable: Yes, It would be difficult to see the difference between toilet with or without drinking water.
Plus, colors begin only an issue in dense area like Paris, but in normal city it's ok.

symbols/openstreetmap-carto/amenity/drinking_water.svg
Florimondable: Well, it's not used, but it's the orginal file so I push it.

PR #457, opened 2020-11-21
Add subicon at POI for bicycle services (pump, tool, rental/repair/retail)

### LINKED ISSUE #445: Render bicycle pump & tools of a POI
In Paris we're having private bicycle parking deployement by the city which have a free public bicycle pump.
I know shops which offer a bicycle pump for free when they are open.
It would be great to have a bicycle pump icon next to the any kind of POI icon in order to give the information.

examples : 
https://www.openstreetmap.org/node/8007638492
https://www.openstreetmap.org/node/7999237089

Florimondable:
Fix #445 (some example in the ticket)

Add rendering for fuel amenities/shop gas.
Render sport shop with bicycle services no more with bicycle shop icon but with sport shop icon and subicons.

Screenshot_20201121_180622
Screenshot_20201121_173159


amenities.mss
```diff
@@ -921,6 +927,46 @@
     marker-fill: @amenity-common;
     marker-clip: false;
   }
+
+  //Pump sub icon
```
Phyks, line 931: Wow, that's a great way of handling it, with subicons! Thanks! :tada:

amenities.mss
```diff
@@ -965,6 +1011,8 @@
   [feature = 'shop_bakery'],
   [feature = 'shop_convenience'],
   [feature = 'shop_convenience;gas'],
+  [feature = 'shop_gas'],
+  [feature = 'amenity_fuel'],
```
Phyks, line 1015: I think we do not want the name of gas / fuel being displayed.

These are usually complex objects in OSM, and I think interesting names are already rendered most of the time, see https://www.cyclosm.org/#map=20/45.89039/4.44568/cyclosm.

Issue #374, opened 2020-05-23
Render flood_prone=yes


treestryder:
I am not sure how this could be implemented without cluttering the map, but it would be nice to have some indication of flood prone areas.

See https://wiki.openstreetmap.org/wiki/Key:flood_prone

Florimondable:
I see, we have the case in Paris (a cycleway and a park used a lot by cyclist).
May be as we do for #65 ?

PR #458, opened 2020-11-25
Render flood_prone=yes + concatenate all conditionals

Phyks:
Closes https://github.com/cyclosm/cyclosm-cartocss-style/issues/374.

I also edited the conditional text to concatenate all available conditionals.

See https://www.openstreetmap.org/way/199162461#map=19/48.81604/2.41142


Florimondable:
We loose the color when conditional is pro cycling.
Not sure it worth it to keep this distinction.

Phyks:
Indeed, we cannot keep the color distinction with the merge of all conditionals :/ I think more conditionals is more useful than the color (e.g. knowing that there won't be any vehicles on this street).

Florimondable:
ok

PR #459, opened 2020-11-25
Truncate long cycle junction names.

### LINKED ISSUE #381: Cycle route node long text bad rendering
See https://www.cyclosm.org/#map=16/48.4666/-4.5631/cyclosm
the name of the node Milizac is render in white and is barely readable.

Phyks:
Fix #381 

See https://www.cyclosm.org/#map=19/48.46640/-4.56664/cyclosm

Render is 

If we want to keep the actual render (which is good, in my opinion), we only have two options:
* Either we respawn our shield building script to generate the background shield in a variety of sizes and formats. Looking at the lcn_ref values from taginfo, I'm definitely not convinced this would be clean enough as the shield names can be arbitrarily long.
* Or we truncate the cycle junction name. My bet here is that non-numeric cycle junction names are likely places names. Hence, given the surroundings in the map, having just a few letters can disambiguate and make a nice render.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Distinguish access for toilets and drinking water.
 * Add subicon at POI for bicycle services (pump, tool, rental/repair/retail).
 * Render islands names. See #456.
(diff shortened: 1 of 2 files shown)


Florimondable:
Ok for me.

PR #461, opened 2020-11-27
Optim time to generate xml due to a regression with subicons

Florimondable:
Bicycle service sub icons are only set for some POI (convenience/supermarket, parking, fuel, repair station, bicycle shop, sport shop).

Was at 57s on my PC, to 14s with this fix.


Phyks:
Thanks!

PR #462, opened 2020-11-27
Mountain pass

### LINKED ISSUE #446: mountain_pass=yes should be rendered with name and elevation
a node with mountain_pass=yes should be shown in cyclosm even if it is not natural=saddle. currently only natural=saddle are shown.
below is an example of a "mountain_pass" which is clearly not a saddle. The information is still useful for a cyclist. 
image

see also #22

https://www.cyclosm.org/#map=18/45.53653/4.75609/cyclosm

Florimondable:
Fix #446, and change zooms for natural peak/voclano and saddle mountain pass.
For cyclist mountain pass (on road or path) is much more important than peaks or saddle (no road).
Hence, mountain pass is rendered from z11 in grey, if a mountain pass is also a saddle, grey is kept.
Peaks and saddles are rendered in higher zooms.

I also clean the sql for bicycle services in amenities text layers.

Screenshot_20201127_194729
Screenshot_20201127_194711
Screenshot_20201127_194632


hiddewie:
Also relevant, https://github.com/gravitystorm/openstreetmap-carto/pull/4121, where I try to get mountain passes to be rendered in the openstreetmap-carto style.

Florimondable:
@hiddewie Thanks, it's interesting.
Wow, it take a long time to have this integrated.

May be you could try as I did, reuse the saddle icon with grey color ?

About the direction of a mountain pass, it's weird unless you have only (and realy only one) road/path (which I think is realy rare) a  mountain pass doesn't have one direction. (This makes me think that by the definition of the wiki, a road can have a mountain pass where as a path can cross this mountain pass  without being a mountain pass for it).

hiddewie:
Yes indeed. The saddle icon could also work. Frankly I do not really mind the icon, I would really like mountain passes to be rendered on the general OSM style.

Let's hope there well be more reactions on the pull request in openstreetmap-carto, and it will be merged in the future.

PR #464, opened 2020-11-28
Render cycleable sidewalks tagged on roads

### LINKED ISSUE #409: Render sidewalk= similar to cycleway=track
In addition to `cycleway=`, which are used for tagging cycle facilities without creating separate ways, there is a similar tag `sidewalk=`. In particular it is the recommended way to tag a sidewalk attached to a road where bicycles are allowed but not designated [according to the German wiki](https://wiki.openstreetmap.org/wiki/DE:Bicycle/Radverkehrsanlagen_kartieren). Unlike `cycleway=`, where the value specifies the type and side is specified as a suffix to the key (e.g. `cycleway:right=track`), `sidewalk=` specifies the side (e.g. `sidewalk=left`).

This could be rendered using borders that match the rendering of `highway=footway` today: a green line if `sidewalk:(<side>:)bicycle=yes`, otherwise brown. These can also be oneway, either for vehicles in general (using `sidewalk:(<side>:)oneway=` or for bicycles specifically using `sidewalk:(<side>:)oneway:bicycle=`, so ideally they would get direction arrows if anything is specified.

There are also cases where one side of a road has both a marked cycle lane (on the road) and a sidewalk where cycling is also permitted (they can also be in opposite directions). Ideally in this case both borders would be rendered, with the bike lane on the inside and the sidewalk on the outside.

brainwad:
Adds two new colours of cycletrack, which represent sidewalks that are primarily for pedestrians but either designated for bicycles (drawn in the same lighter blue as a footway designated for bicycles is) or merely allowed for bicycles (drawn in the same green as a footway that allows bicycles).

image
image

The ranking of road-based bicycle features that are drawn in the border of the road is now:

A cycletrack (`cycleway=track`) or a segregated sidewalk, that allows bicycles (`sidewalk:bicycle!=no sidewalk:segregated=yes`): solid dark blue
A sidewalk designated for bicycles (`sidewalk:bicycle=designated`): solid lighter blue
A bike lane (`cycleway=lane`): dashed dark blue
A sidewalk allowing bicycles (`sidewalk:bicycle=yes`): solid green
A shared busway or lane: very dashed dark blue

The reasoning for splitting the sidewalks is that a designated off-road path is probably more comfortable than an on-road lane, but a non-designated sidewalk can potentially quite bad infrastructure and so a dedicated lane is better; but then sharing with pedestrians is still better than sharing with cars, so shared road lanes come after it.

Fixes #409.


Florimondable:
There is 7 sidewalk:bicycle=designated don't worth a special color for it.

And 679 for sidewalk:bicycle, does it worth the effort and the rendering cost ?

https://taginfo.openstreetmap.org/keys/sidewalk%3Abicycle#values

brainwad:
sidewalk:bicycle isn't a valid tagging. Look at
sidewalk:(both/left/right):bicycle

Phyks:
> sidewalk:bicycle isn't a valid tagging, that's why its so uncommon (and I don't even bother to render it). Look at sidewalk:(both/left/right):bicycle : https://overpass-turbo.eu/s/10Fl

There are around 10k `sidewalk:[left/right/both]:bicycle=yes ` tags around, this is definitely something worth rendering.

However, there are only 32 `sidewalk:both:bicycle=designated` + 292 (left case) + 416 (right case) which is < 1k objects worldwide.

Given the tag adoption, I think it's not worth making the distinction (perf-wise). I would simply check the value against anything but `no` and `dismount` and have a single "sidewalk" render.

project.mml
```diff
@@ -675,10 +683,16 @@ Layer:
           END AS oneway,
           CASE
             WHEN tags->'cycleway:left' IN ('track', 'opposite_track') THEN 'track'
+            WHEN tags->'sidewalk:left:bicycle' <> 'no' and tags->'sidewalk:left:segregated' = 'yes' then 'track'
```
Phyks, line 686: Minor (stylistic) comment but could we have an homogeneous SQL code convention here?

Almost all of CyclOSM SQL queries are written using `!=` operator (instead of `<>`) and uppercase keywords (`AND` / `THEN`). Thanks!

Phyks:
I have a SQL query style comment, and I'd like to drop the dedicated `sidewalk:*:bicycle=designated` render. Once done, I think it is a go for merge.

brainwad:
SGTM, in that case I will rank sidewalk between lane and shared lane.

Florimondable:
> sidewalk:bicycle isn't a valid tagging, that's why its so uncommon (and I don't even bother to render it). Look at sidewalk:(both/left/right):bicycle : https://overpass-turbo.eu/s/10Fl

My bad.
So yes let's try with a "solid lighter blue".

brainwad:
I think path green would be better, since most usages are not designated, so having the same colour as `highway=footway bicycle=yes` makes more sense. As a bonus, it's more distinguishable from cycletracks when its green.

Florimondable:
ok

brainwad:
The area above, with both designated and non-designated cycleable sidewalks now looks like this:

image

Phyks:
Thanks!

PR #465, opened 2020-11-29
Set up CI

Phyks:
WIP


Phyks:
Mapnik XML file will now be built for each commit / pull request. It helps ensuring nothing is broken (for the Mapnik XML generation, it does not test rendering against a database) and might help keeping track of build times.

PR #466, opened 2020-12-02
Add cycle route name label

### LINKED ISSUE #433: Draw cycle route full name along way at high zoom
Because just there codename is not very friendly and are confusing.

Florimondable:
Render next to the road the name of the cycle route with the network color darkened.
From z16.
Shield are more spaced to compensate the density.

fix #433

Screenshot_20201202_163207


Phyks:
:heart_eyes:

Issue #442, opened 2020-10-20
Closed linear leisure=track+sport=cycling ways are rendered as an area


Florimondable:
see :
https://www.openstreetmap.org/way/158239095
https://www.cyclosm.org/#map=19/48.82535/2.48268/cyclosm

Florimondable:
But this one is correctly rendered https://www.openstreetmap.org/way/28836778
It has area=no, but closed leisure=track way should be treated as linear by default.

quasart:
according to [osm wiki](https://wiki.openstreetmap.org/wiki/Tag:leisure=track?uselang=en) :

> It's unclear/undecided whether closed ways should be treated as areas for this tag so it is advised to always add an area=yes/no tag in the case of closed ways.

Florimondable:
and « The default OSM Carto stylesheet displays an area only if area=yes is present. »
We should do the same

Florimondable:
So after looking it up, it seems that by default osm2pgsql create a polygon for leisure tagged closed way. OSM Carto has a osm2pgsql import script which create an exception for leisure=track|slipway.

So I guess we will not change this default behaviour (otherwise we have to reimport the whole OSM world).

Phyks:
Same, I guess we can close this one.

PR #472, opened 2020-12-12
Render leisure track bicycle for areas

### LINKED ISSUE #440: Render leisure=track+sport=cycling area like the linear ones
see #412

Florimondable:
Fix #440 and #442

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add cycle route name.
 * Render parking amenity and servie road parking aisle, drive-through, driveway smaller.
 * Add floor/ground slots bicycle parking.
+* Add bicycle leisure track for areas.
 
 
 ## v0.3.7
--- a/base.mss
+++ b/base.mss
@@ -152,27 +152,38 @@
     polygon-fill: @track;
     line-width: 0.5;
     line-color: @track * 0.95;
+
+    [sport='cycling'][zoom >= 15],
+    [sport='bmx'][zoom >= 15] {
+      line-cap: round;
+      line-join: round;
+      line-color: @bicycle-amenity;
+      line-width: 0.2 * @rdz15_cycle;
+      [zoom>=16] { line-width: 0.4 * @rdz16_cycle; }
+      [zoom>=17] { line-width: 0.5 * @rdz17_cycle; }
+      [zoom>=18] { line-width: 0.6 * @rdz18_cycle; }
+    }
   }
   [type='amenity_university']    { polygon-fill: @school; }
   [type='natural_wood']          { polygon-fill: @wooded; }
 }
 
-/* Display linear leisure_track like an area. */
+//linear leisure track
 #leisure_track[zoom >= 11] {
-  ::area_ouline {
-    line-cap: round;
-    line-join: round;
-    line-color: @track * 0.95;
-    line-width: 1 + @rdz12_track;
-    [zoom>=13] { line-width: 1 + @rdz13_track; }
-    [zoom>=14] { line-width: 1 + @rdz14_track; }
-    [zoom>=15] { line-width: 1 + @rdz15_track; }
-    [zoom>=16] { line-width: 1 + @rdz16_track; }
-    [zoom>=17] { line-width: 1 + @rdz17_track; }
-    [zoom>=18] { line-width: 1 + @rdz18_track; }
-  }
-
   ::area {
+    comp-op: darken;
+
+    outline/line-cap: round;
+    outline/line-join: round;
+    outline/line-color: @track * 0.95;
+    outline/line-width: 1 + @rdz12_track;
+    [zoom>=13] { outline/line-width: 1 + @rdz13_track; }
+    [zoom>=14] { outline/line-width: 1 + @rdz14_track; }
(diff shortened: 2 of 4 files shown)


quasart:
Hello !

In this fix, area borders and way tracks are rendered the same way, I find this confusing. 
In the Velodrome Anquetil example, we may understand there are two concentric tracks.

Here is a try with an area fill :
Screen Shot

And in your bicross example, I understand it's actually not an area. Rendering would already look nice with explicit "area=no" tagging (advised tagging in OSM wiki)

Cheers

Florimondable:
> Hello !
> 
> In this fix, area borders and way tracks are rendered the same way,
Yeah, that's the trick so area or linear ways can be mixed properly.

Phyks:
What about https://postgis.net/docs/ST_ExteriorRing.html? (Performance should be tested)

Florimondable:
> What about https://postgis.net/docs/ST_ExteriorRing.html? (Performance should be tested)

Not convinced by the rendering result it'd make.

Phyks:
In the case of https://user-images.githubusercontent.com/47089717/101991384-a7ab6080-3cac-11eb-8582-12ed46b5673b.png, it would return only the outer ring (which is the outer purple line). Then, it should discard the misleading inner loop.

Florimondable:
I don't fill it's that much missleading.

quasart:
> Yeah, that's the trick so area or linear ways can be mixed properly.

I understand the trick, it’s a smart try but i think It’s not worth the confusion it creates :
- it may mislead the map reader
- For developpers, it hard to choose appropriate rendering style. In your example violet casing disappear at z14 : this is good for #442 but bad for #440

I think KISS principle Is better here : Bold purpul for lines, and lighter purpul for areas

Florimondable:
Let me explain the problem again :
we have leisure=track closed way which is considered polygon by osm2pgsql, where as I consider as OSM Carto they are linear way.
I cannot change current osm2pgsql conf.

So I had to find a middle ground where both areas and linear leisure=track are well drawn.

quasart:
I understood that very well. I guess my answer is not clear.

Since we cannot know if a polygon is an actual area or a closed I see two solution :
1. a "middle ground" rendering style looking like both an area and a track
2. always render it as an area

I'm saying (1) is confusing for both map reader and style developer.

Solution (2) is KISS and we still may advise OSM contributor to add explicit `area=no` if they want clean rendering. It's recommended tagging anyway.

Florimondable:
@Phyks Can I merge it, so we can move forward ?

Phyks:
Hi,

Sorry, this one slipped out of my sight :/ I'm still a bit puzzled by the rendering, which seems a bit misleading for me. From the OSM wiki, I see

> It's unclear/undecided whether closed ways should be treated as areas for this tag so it is advised to always add an area=yes/no tag in the case of closed ways. The default OSM Carto stylesheet displays an area only if area=yes is present. This is not necessary for multipolygons since they are always handled as areas. 

I think OSM carto approach is nice (requiring an explicit `area=yes`) but this needs tweaking osm2pgsql, which we don't want to do. I, however, think we can achieve the same thing by tweaking SQL queries. That is:

The current `leisure_track` is enough to target **only** the linear `leisure=track`:
```
- id: leisure_track
  <<: *extents
  Datasource:
    <<: *osm2pgsql
    table: |-
      (
        SELECT
          way,
          sport
        FROM planet_osm_line
        WHERE leisure = 'track'
      ) AS data
  geometry: linestring
  properties:
    cache-features: true
    minzoom: 11
``` 

We have to add rendering for areas `leisure=track`, without perturbing this one. We have to :
- include those with an explicit `area=yes`, e.g. https://www.openstreetmap.org/way/787733437
- include those which come from a `multipolygon` relation, as they are to be treated as areas by default, e.g. https://www.openstreetmap.org/relation/3214796
- exclude any closed way without explicit `area=yes`, e.g. https://www.openstreetmap.org/way/158239095

This can be done with the following SQL query:

```sql
SELECT
    osm_id,
    leisure
FROM (
    -- Distinguish two cases: either there is an explicit area=yes or this is coming from a realtion with multipolygon type
    (
        SELECT
            planet_osm_polygon.osm_id,
            planet_osm_polygon.leisure,
            true AS is_area
        FROM planet_osm_polygon
        WHERE
            -- Fixed input list of osm_id for testing. Should be way && !bbox! in production
            osm_id IN (787733437, -3214796, 158239095)
            AND area='yes'
    )
    UNION
    (
        SELECT
            planet_osm_polygon.osm_id,
            planet_osm_polygon.leisure,
            planet_osm_rels.tags @> '{type, multipolygon}' AS is_area
        -- osm_id handling is different across these two tables, hence the need for -1 multiplication
        FROM planet_osm_polygon LEFT JOIN planet_osm_rels ON planet_osm_polygon.osm_id = - planet_osm_rels.id
        WHERE
            -- Fixed input list of osm_id for testing. Should be way && !bbox! in production
            osm_id IN (787733437, -3214796, 158239095)
            AND osm_id < 0
            AND (area IS NULL OR area != 'yes')
    )
) AS data;
```

This is quite slow at the moment due to the osm_id trick for relations. I guess this could be achievable in production by adding an index on this, to be checked.

This can be quite heavily simplified if we only consider explicit `area=yes` ways (even for multipolygons).

Florimondable:
So yeah, it's over complicated for not much in my opinion.

PR #476, opened 2020-12-18
Add xmas tree and market at the end of December.

Florimondable:
I had time to loose 🎅

Only rendered after 15 of December.

image

### DIFF
--- a/amenities.mss
+++ b/amenities.mss
@@ -1186,6 +1186,18 @@
     marker-fill: @amenity-common;
     marker-clip: false;
   }
+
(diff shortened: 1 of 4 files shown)


Florimondable:
May be we can set the date earlier.

Phyks:
🎉 

But I think you forgot the svg files :)

Florimondable:
Here it is.
(forgotten as always :sweat: )

Phyks:
This is now in production along with v0.4! \o/

Issue #28, opened 2019-01-29
Construction


Florimondable:
Draw construction highway, and firstly cycleways.

A faded version of the normal style of the highway would be enough I think.

Phyks:
Example of construction highway (for next year at least) is https://www.openstreetmap.org/way/224483432. Which is currently rendered as https://tiles.phyks.me/#18/48.81797/2.31829 (small line).

Do you have an example with cycleways?

Florimondable:
Alright I didn't saw them.
You have cycleway example on the famous Champs Élysée ;)
https://tiles.phyks.me/#18/48.86994/2.30738

So the ticket would changed to : better visualization, I can't barely identify them (which type of higway ?), I like the default OSM render like https://www.openstreetmap.org/#map=19/48.86968/2.30849

Phyks:
What's the status of this one? Do we have an updated example at hand?

Florimondable:
Current status is that construction roads are not rendered
Cycleway example here https://www.cyclosm.org/#map=19/48.84389/2.37202/openstreetmap

Phyks:
Got an idea on this, we could transform `highway=construction` + `construction=?` into `highway=?` + `construction=true` in `project.mml`.

Then, the highway under construction would get the exact same render as the final one. Then, we need to distinguish the construction status of the way.

What about using transparency and dimming it out, just as we do for `private` POIs?

Florimondable:
Dash line ? Transparency and dash line ?

PR #485, opened 2021-01-03
Render highway=construction

Phyks:
Hi,

Here is a first draft of render for `highway=construction` tag as discussed in #28. I'm only playing on opacity so far, since playing with dash would make it difficult to read and risk inducing confusion with the already dashed items (cycleway, footway, etc).

Some examples below:

https://www.openstreetmap.org/way/667390515#map=19/45.77706/3.07498

https://www.openstreetmap.org/way/854996738#map=17/46.14530/3.42974

Render is not perfect, I identified two issues:

1. Playing with opacity makes the road appear grey-er than it should (due to opacity not being applied to the whole object but layer by layer and hence grey is made of semi-opaque white on top of grey outline).
2. Junctions are not rendered nicely (see roundabout).

This however does seem fine to me for a first render and convey the necessary information about "this is a construction area, you might or not be able to cross".

Best,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Render `highway=construction`. See #28.
+
 
 ## v0.4
 
--- a/project.mml
+++ b/project.mml
@@ -451,6 +451,7 @@ Layer:
           way,
           COALESCE(
             CASE
+              WHEN highway='construction' THEN construction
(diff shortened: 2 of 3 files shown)


jeisenbe:
I would caution against use of opacity, if possible. Since there are different background colors created by different types of land and buildings, using opacity will cause the colors to mix. When there are two different hues this can have unexpected results and make the map less legible. 

If you use grayscale the problem is not as severe, but it can still be harder to understand for map users.

Phyks:
> I would caution against use of opacity, if possible.

Indeed, this is a source of problems.

@jeisenbe do you have any other rendering ideas for such `highway=construction` to convey the same meaning of "this road does not really exist"?

PR #499, opened 2021-01-20
Add dashline outline on path like way tunnels

### LINKED ISSUE #492: Hard to see tunnels: cycleways on different levels look like they cross each other
Where a cycleway passes under another through a tunnel, the rendering doesn't clearly show that there is a tunnel. This makes it look as if the cycleways cross instead of one going underneath the other.

Example ([OSM](https://www.openstreetmap.org/#map=19/53.19189/5.78003&layers=N)):

Screenshot from 2021-01-16 14-19-35

It does look like tunnels are in a lighter blue, but I find it hard to tell.

OSM-Carto for comparison:

Screenshot from 2021-01-16 14-19-46

-----

Nice to see the CyclOSM layer on openstreetmap.org as well. I find it very useful as a mapper to quickly see different renderings styles for different purposes after working on an area. :+1:

Florimondable:
#fix #492 

Screenshot_20210120_230251
Screenshot_20210120_230230
Screenshot_20210120_230156

The outline for bridges is also little wider.
Screenshot_20210120_230030

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 * Render more features at low zoom: land, water and hillshades at zoom 3 and 4. See #494.
+* Add dashline outline on path like way tunnels.
 
 ## v0.4
 
--- a/road-colors.mss
+++ b/road-colors.mss
@@ -54,6 +54,7 @@
 @track-light2: lighten(@track-fill, 50);
 
 @bridge_case: #666;
+@bridge_path_case: @land;
 
 @aeroway: #ddd;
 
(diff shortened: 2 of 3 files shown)


Florimondable:
Add dashline for path and bridleway, here a case with a segregated path :
Screenshot_20210120_231721


Remove dashline for footway, otherwise it would be too much in city :
Screenshot_20210120_231926

Phyks:
What about making tunnel blue lighter as well?

Florimondable:
> What about making tunnel blue lighter as well?

It's already slightly lighter

Phyks:
It is, but I would make the contrast a bit clearer. This would help distinguish the two blue tones even on bad quality screens (smartphones for instance).

Florimondable:
ok,
Screenshot_20210122_155900
Screenshot_20210122_155819
Screenshot_20210122_155737

Phyks:
Thanks!

PR #507, opened 2021-01-25
Fix udnerground bicycle parkings masking buildings. Fix #505.

### LINKED ISSUE #505: Underground parking amenities are rendered on top of ways
In the Netherlands underground bicycle parking facilities are not uncommon near larger train stations. [One such example is in Leeuwarden](https://www.openstreetmap.org/way/518639356).

In CyclOSM this `amenity=bicycle_parking` tagged with `location=underground` and `layer=-1` is drawn on top of the ways above ground:

Screenshot from 2021-01-23 18-34-54

It is also rendered on top of the building above it.

There may be two bugs here:

* Objects should not be rendered on top of those with a higher layer (here cycleways and building with the default `layer=0`).
* Underground objects (`location=underground`) should be rendered differently or not at all (although their amenity might still show up as a node).

Phyks:


Florimondable:
Well, I'd like to render also underground bicycle parking, no ?

Phyks:
It's still rendered, but only with an icon, not with the whole surface render.

Florimondable:
Ok, but I think it's usefull to know what is the shape of the bicycle parking to know where is the entrance. An icon lost in nowhere doesn't help much.

May be we should render amenities_poly under road layers ? With building layer ?

Phyks:
We cannot move the whole `amenities-poly` layer since we want the icons to be always on top.

We might move the surface rendering in the `landuse` layer? https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/project.mml#L137-L165

This would not work for a surface bicycle parking on top of the roof of a building though.

Florimondable:
> We cannot move the whole `amenities-poly` layer since we want the icons to be always on top.
Yes, we need to keep this layer for icon rendering, but area can be moved.
> 
> We might move the surface rendering in the `landuse` layer? https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/project.mml#L137-L165
> 
> This would not work for a surface bicycle parking on top of the roof of a building though.
Yeah, I think we need to think at "how to render bicycle parking area as an indoor, or building like amenities", so may be we can use the "buildings" layer ?
Roads would be rendered over, and we could layers things over/under the bicycle parking areas.

PR #531, opened 2021-03-14
Improve clpc rendering

### LINKED ISSUE #528: Give way for cyclist arrows are difficult to read
Coming from the north you can go west (turn right).
Coming from the west you can go south (turn right).
The arrows are in the point of view of the cyclist "turn right" so we draw an arrow going to the right.
But with the point of view of the map reader, it would be easier to read if the arrow is pointing toward the direction on the map.
In this example with arrow going west (left) and south (down).

https://www.cyclosm.org/#map=20/48.87775/2.36550/cyclosm_live
image

Phyks:
Fix #528

https://www.cyclosm.org/#map=20/48.87775/2.36550/cyclosm_live
**Was: **
image

**Becomes:**
Screenshot_2021-03-14 CyclOSM — Kosmtik

https://www.cyclosm.org/#map=20/48.83984/2.34750/cyclosm_live
**Was:**
image

**Becomes:**
Screenshot_2021-03-14 CyclOSM — Kosmtik2

### DIFF
--- a/amenities.mss
+++ b/amenities.mss
@@ -1565,12 +1565,12 @@
 #clpc[zoom >= 19] {
     marker-width: 20;
     marker-file: url('symbols/clpc/right.svg');
-    [angle <= 45],
-    [angle >= 135][angle <= 225] {
+
+    [from_to_angle <= 45],
+    [from_to_angle >= 135][from_to_angle <= 225] {
       marker-file: url('symbols/clpc/straight.svg');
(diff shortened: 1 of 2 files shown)


Florimondable:
Nice, it's way better.
Don't you want to try to render the board (triangle+arrow) on the way ?

Phyks:
Indeed, you are right. Rendering is way better with signs on the ways and code is way simpler!

Florimondable:
Nice, it's better I think

PR #537, opened 2021-03-21
Render landuse=retail and shop=mall names

### LINKED ISSUE #526: Commercial areas are not easy to spot
See https://www.openstreetmap.org/#map=16/48.7818/2.2180

Phyks:
Fix #526.

https://www.openstreetmap.org/#map=16/45.6480/-1.0490

https://www.openstreetmap.org/#map=16/48.7818/2.2229


Florimondable:
I thought you wanted to change the landuse background color ?

Phyks:
Name is enough for my use case described in related issue. I'm open to change landuse background as well, but this does not seem to be necessary for me.

Florimondable:
```
@commercial:        @land * 0.97;
@industrial:        @land * 0.96;
```
Hard to see the differences ':).

I'd be ok to change the color to have something different, maybe grey or the red color like osm-carto ? Commercial areas is much more usefull to know than industrial areas, I think.

Phyks:
Even more since it was not used https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/base.mss#L36-L39 :)

Fixed in another commit. See https://www.openstreetmap.org/#map=17/48.77426/2.22357

retail

Florimondable:
Ok, I think the red conflict too much with the military landuse color.

I tried, light purple #ded8dd
Screenshot_20210322_220000

And a grey #d6d6d6
Screenshot_20210322_220251

I think both work, what do you think ?

Phyks:
Indeed! Slight preference for the light purple on my side, which seems to convey a bit more semantic than just plain grey. PR is updated with conflict resolution and light purple.

quasart:
Hello, I think it's good idea to use a light color instead of grey, but why purple ?

I though purple means cycling-related amenities ? (cycling track, routes, bicycle shops, compressed air)

Why not try light brown instead ?

PR #538, opened 2021-03-21
Render defibrillators

### LINKED ISSUE #496: Defibrillators, water
A lot of first aid/emergency features are included.  Given that, the lack of emergency=defibrillator would appear to be an omission.

Also it's nice to see water points included.  Perhaps the documentation should list which tags are included in this, as there are several of interest

Phyks:
Fix #496

Render defibrillators from Z19, just the same way as `emergency=phone` were rendered.

aed

### DIFF
--- a/amenities.mss
+++ b/amenities.mss
@@ -465,6 +465,14 @@
     marker-clip: false;
   }
 
+  [feature = 'emergency_defibrillator'][zoom >= 19] {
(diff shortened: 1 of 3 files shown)


Florimondable:
Ok, but I don't think it's link to cycling, no ?

Phyks:
Not so much, but it goes together with the rest of emergency features which can be useful while cycling: pharmacies, emergency phones, hospitals etc.

PR #539, opened 2021-03-21
Add a bike bottle subicon for drinking_water:refill=yes

### LINKED ISSUE #325: Render drinking_water:refill
drinking_water:refill is a tag that was recently approved to indicate if the establishment participates in a water refill network. The tag is currently being used on less than 100 features but hopefully it will gain traction in the coming months. I feel like the tag is very useful for cyclists in counties, such as the UK, where there are not many water fountains but there are many establishments participating in bottle refill schemes.

Generally features tagged with drinking_water:refill=yes are going to be cafes, bars, restaurants and shops. 

I'm unsure of a suitable way to render the tag, maybe cafes, bars, restaurants and shops could be blue instead of the standard grey if they provide refills?

Phyks:
See "La Succulente" here : https://www.openstreetmap.org/#map=19/43.67937/4.63147

Fix #325.


Florimondable:
Remarks :
- Only 400 objects is tagged with drinking_water:refill=yes, that's not really much.
- The icon looks like a piece of a puzzle, may be we could just use the water glass of the drinking water icon.
- The icon should use the same blue as drinking water.

Phyks:
This is only 400 items, but there seems to be some traction around this tag in the OSM community at the moment.

Here is a fix for the icon + color:

drinking_water_refill

Phyks:
https://taginfo.openstreetmap.org/keys/drinking_water%3Arefill#chronology not enough traction. Closing

Issue #552, opened 2021-04-14
Rendering add : waterway=dock


CorentinLemaitre:
Hello,

First 👏 for the quality of the map for bicycle ! Thanks for that. 

I have seen that the tag waterway=dock is not rendered in the cyclosm map. Here is an example : [LINK](https://www.openstreetmap.org/#map=15/54.1024/-3.2217&layers=Y).

Here is the wiki description [LINK](https://wiki.openstreetmap.org/wiki/Tag:waterway%3Ddock)

PR #553, opened 2021-04-14
Add waterway = dock

CorentinLemaitre:
I did use the same filter for the riverbank adding only dock.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -332,7 +332,7 @@ Layer:
         WHERE
           (
             "natural" IN ('water', 'pond')
-            OR waterway IN ('basin', 'canal', 'mill_pond', 'pond', 'riverbank', 'stream')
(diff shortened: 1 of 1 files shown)


CorentinLemaitre:
I have use the same rule used for the waterway = riverbank to adapt it to waterway = dock

CorentinLemaitre:
Related to issue #552

Phyks:
Thanks for the catch and the PR! It will be part of next version.

CorentinLemaitre:
I see no movment in the style produced online for region where there is dock ([like here](https://www.cyclosm.org/#map=16/47.2761/-2.1986/https://www.cyclosm.org/#map=16/47.2761/-2.1986/))

It is just to update it as I don't know if there is still time to wait for the style to be change online.

Florimondable:
It’s not yet into production

PR #558, opened 2021-04-24
remove sql code duplication

quasart:
Hello,

I've noticed sql code duplication in project.mml, 
I think it's better to factor this code by calling a common sql view. This code factoring should avoid some bugs, and would make future dev and rework easier.

This PR is a POC for roads. I think it could be done for other layers, for instance routes.

What do you think ?


Phyks:
Hi,

Thanks a lot, huge +1 for deduplication here. The same stands for amenities. 

Not sure whether we could achieve the dame with yaml variables? Otherwise, psql views are probably fine, i'll just have to quickly double check with osm-fr which is hosting us and providing us with a planet db.

Phyks:
I double checked, it's fine to use SQL views as proposed here!

quasart:
I eventually couldn’t find nice rework of routes. Do you think other layers should be handled with views ?

Phyks:
It can probably be done as well for amenities-{points,poly} https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/project.mml#L1763-L2066 and amenities-{points,poly}-text https://github.com/cyclosm/cyclosm-cartocss-style/blob/master/project.mml#L2205-L2523

quasart:
Oh this was done already in 5dfe79d

Phyks:
Indeed, I missed this extra commit, sorry!

Unless you find other areas of interest for SQL code deduplication (I don't have any in mind), I think this one is ready for review. Would you want to fix conflicts or do you prefer I handle it upon review / merge?

quasart:
Conflicts fixed :heavy_check_mark:

Phyks:
Great, thanks!

Issue #557, opened 2021-04-23
Add rendering for `highway=busway`


kolgza:
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbusway

Florimondable:
Should be rendered like service or unclassified/residential road.
To be merged in the sql request.

PR #587, opened 2021-07-04
Proposal for a fix for highway=busway #557

Famlam:
Note: I have very little knowledge of SQL, so I might not be able to include all feedback, but hopefully this is a good start?
(Based on the way `highway=raceway` was added, plus how `motorway`s are by default (just like `busway`) inaccessible for cyclists)


Phyks:
Thanks! (and sorry for the delay for the review!)

Issue #601, opened 2021-10-05
Cycleway with no oneway=no is drawn with a the width of a oneway=yes


Florimondable:
See the small width of this cycleway without `oneway` tag https://www.openstreetmap.org/way/265704255#map=19/48.83451/2.26533&layers=Y

vs the width of this `oneway=no` https://www.openstreetmap.org/way/834107886#map=19/48.83468/2.26540&layers=Y

examples around this place https://www.cyclosm.org/#map=20/48.83465/2.26594/cyclosm

Expected result : cycleway with no `oneway` tag should be drawn like a cycleway with `oneway=no` tag.

PR #627, opened 2022-07-01
Add fallback values for oneway_bicycle to views.sql

kaneap:
Cycleways and paths without a oneway tag will now be rendered identically a path/cycleway with a oneway=no tag. Fxes #601

The following screenshot from around https://www.cyclosm.org/#map=18/50.73743/7.07980/cyclosm
Before:
Before
After:
After

### DIFF
--- a/views.sql
+++ b/views.sql
@@ -49,7 +49,7 @@ CREATE VIEW cyclosm_ways AS
         END AS cyclestreet,
         CASE
             WHEN oneway IN ('yes', '-1') THEN oneway
-            WHEN junction IN ('roundabout') AND (oneway IS NULL OR NOT oneway IN ('no', 'reversible')) THEN 'yes'
+            WHEN junction IN ('roundabout') AND (oneway IS NULL OR oneway NOT IN ('no', 'reversible')) THEN 'yes'
             ELSE 'no'
         END AS oneway,
         CASE
--- a/views.sql
+++ b/views.sql
@@ -120,7 +120,7 @@ CREATE VIEW cyclosm_ways AS
               OR tags->'cycleway:left:oneway'='-1' OR tags->'cycleway:right:oneway'='-1'
                 THEN 'no'
             WHEN oneway IN ('yes', '-1') THEN oneway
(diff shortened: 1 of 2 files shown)


views.sql
```diff
@@ -113,14 +113,15 @@ CREATE VIEW cyclosm_ways AS
...
             WHEN tags->'oneway:bicycle' IS NOT NULL THEN tags->'oneway:bicycle'
-            WHEN highway='cycleway' AND oneway IS NOT NULL THEN oneway
             WHEN tags->'cycleway' IN ('opposite', 'opposite_lane', 'opposite_track', 'opposite_share_busway')
               OR tags->'cycleway:both' IN ('opposite', 'opposite_lane', 'opposite_track', 'opposite_share_busway')
               OR tags->'cycleway:left' IN ('opposite', 'opposite_lane', 'opposite_track', 'opposite_share_busway')
               OR tags->'cycleway:right' IN ('opposite', 'opposite_lane', 'opposite_track', 'opposite_share_busway')
               OR tags->'cycleway:left:oneway'='-1' OR tags->'cycleway:right:oneway'='-1'
                 THEN 'no'
-            ELSE NULL
+            WHEN oneway IN ('yes', '-1') THEN oneway
```
Phyks, line 122: The changes go beyond the description of this MR (defaulting `highway=cycleway` without `oneway` tag to `oneway=no`).

The global setting is changed (`NULL` => `no`) and there is added logic for junctions.

I'm not sure to see why this is needed?

Thanks

Phyks (commented):
Sorry, I've just realized Github stopped sending me email notifications a while ago and I am going over the backlog. Not 100% sure to understand why so much changes in oneway logic is required?

views.sql
kaneap: The idea is that the cycleways/paths are rendered with the correct width when the implied oneway:bicycle=* value is present. This logic simply applies the fallback oneway:bicycle=* value in the view when it is not tagged explicitly.

Phyks: Ok got it, thanks for the fix.

PR #652, opened 2023-07-06
Render unsegregated cycle track as light blue

### LINKED ISSUE #640: Proposal: render cycleway:[left:|right:|both:]segregated:no similar to shared bicycle/foot path
There are cycling lanes which are at the same time used as a sidewalk, like here: https://www.cyclosm.org/#map=19/47.43148/9.66819/cyclosm (tagged with cycleway:right:segregated=no and sidewalk:right:segregated=no). I suggest that in such a case, the lane should be marked with the lighter blue, much like unsegregated shared bicycle/foot paths.

kaneap:
Fixes #640

### Description
This PR sets the color of `cycleway(:left/:right/:both)=track` to light blue (same color as `bicycle=designated` on unsegregated `highway=path`) when `cycleway(:left/:right/:both):segregated=no`.
The rules xml file generated is slightly larger after this change: From 7MB to 7.4MB

#### Example 1:
A two-way shared cycle and foot path on one side of the road.
Location: https://www.openstreetmap.org/way/50741273
Before:
shared_path_before
After:
shared_path_after

#### Example 2:
A one-way shared cycle and foot path on one side of the road.
Location: https://www.openstreetmap.org/way/654562266
Before:
shared_2_before
After:
shared_2_after

#### Example 3:
One way shared cycle and foot paths on both sides of the road.
Location: https://www.openstreetmap.org/way/515092494
Before:
shared_3_before
After:
shared_3_after


Phyks:
Hi!

Thanks for the PR! I did not do a full review yet, but looks good to me. I'll have a deeper look and merge ASAP.

Phyks:
Merged, thanks! It will be available in the next deployed version.

PR #720, opened 2026-06-13
Modernize Docker setup

thof:
CyclOSM is my go-to map when it comes to planning bike routes but unfortunately the openstreetmap.fr tile server is often sluggish to respond or doesn't work at all. So the main goal of this change is to allow anyone to run their own tile server using Docker/Podman:
* Modernize Docker images to support both amd64 and arm64 (I tested the full pipeline on Apple Silicon and Linux box with arm64 CPUs)
* Add a pipeline for pre-rendering raster tiles (still a resource-intensive process)
* Add support for Podman

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,29 @@
-FROM ubuntu:bionic
+FROM debian:trixie-slim
 
-# Style dependencies
+# Style + kosmtik dependencies. git builds kosmtik from sources below.
+# Mapnik itself comes from @mapnik/mapnik's prebuilt core package (pulled by
+# npm), so no system Mapnik / build toolchain is required.
 RUN apt-get update && apt-get install --no-install-recommends -y \
-    ca-certificates curl gnupg postgresql-client python fonts-hanazono \
-    fonts-noto-cjk fonts-noto-hinted fonts-noto-unhinted mapnik-utils \
-    nodejs npm ttf-unifont unzip && rm -rf /var/lib/apt/lists/*
+    ca-certificates curl git python3 postgresql-client sqlite3 nodejs npm \
+    unzip fonts-hanazono fonts-noto-cjk fonts-noto-core fonts-noto-extra \
+    fonts-unifont && rm -rf /var/lib/apt/lists/*
 
-# Kosmtik with plugins, forcing prefix to /usr because bionic sets
-# npm prefix to /usr/local, which breaks the install
-RUN npm set prefix /usr && npm install -g kosmtik
+# @mapnik/mapnik ships a prebuilt core, but its preinstall script aborts under
+# `set -e` when `mapnik-config -v` is missing (Mapnik 4 dropped mapnik-config).
+# A tiny shim answering -v lets preinstall pass; the prebuilt core is then used
+# as-is with no compilation.
+RUN printf '#!/bin/sh\ncase "$1" in -v|--version) echo 4.2.2;; *) echo;; esac\n' \
+    > /usr/local/bin/mapnik-config && chmod +x /usr/local/bin/mapnik-config
 
-WORKDIR /usr/lib/node_modules/kosmtik/
+# Build kosmtik from sources (latest master)
+# @mapnik/core hoisted as a sibling so node-gyp-build picks up the prebuilt
+# binary instead of trying to compile; a wrapper exposes the kosmtik CLI.
+RUN git clone --depth 1 https://github.com/kosmtik/kosmtik.git /opt/kosmtik \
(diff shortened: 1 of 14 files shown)


docker-compose.podman.yml
```diff
@@ -0,0 +1,22 @@
...
+#
+# Why this is needed: the `kosmtik` and `render` containers run as USER 1000 in
+# the image, but must WRITE to the bind-mounted repo (tmp/, data/, mbtiles/).
+# Under rootless Podman your host user maps to container UID 0, so we run these
+# two services as root inside the container — that maps back to your host user
+# and can write host-owned files.
+#
+# We use `user: 0` rather than `userns_mode: keep-id` because podman-compose puts
+# services in a shared pod by default, and Podman rejects a per-container
+# --userns together with --pod ("--userns and --pod cannot be set together").
```
Phyks, line 13: Is the rootless podman setup tested on your end? You mention having tested the full pipeline on multiple machines, but Docker only.

Dockerfile.db
```diff
@@ -1,3 +1,3 @@
-FROM mdillon/postgis
```
Phyks, line 1: Why not https://hub.docker.com/r/postgis/postgis?

Dockerfile
```diff
@@ -1,16 +1,29 @@
...
 
-# Kosmtik with plugins, forcing prefix to /usr because bionic sets
-# npm prefix to /usr/local, which breaks the install
-RUN npm set prefix /usr && npm install -g kosmtik
+# @mapnik/mapnik ships a prebuilt core, but its preinstall script aborts under
+# `set -e` when `mapnik-config -v` is missing (Mapnik 4 dropped mapnik-config).
+# A tiny shim answering -v lets preinstall pass; the prebuilt core is then used
+# as-is with no compilation.
+RUN printf '#!/bin/sh\ncase "$1" in -v|--version) echo 4.2.2;; *) echo;; esac\n' \
+    > /usr/local/bin/mapnik-config && chmod +x /usr/local/bin/mapnik-config
```
Phyks, line 16: Not sure to understand how it should build / install without a proper local mapnik core install? https://github.com/mapnik/node-mapnik/blob/master/scripts/preinstall.sh#L10-L13

Anyways, I'd be more incline to disable/enable set -e before and after the step, rather than lie on the mapnik-config.

scripts/render-tiles.js
```diff
@@ -0,0 +1,267 @@
+#!/usr/bin/env node
```
Phyks, line 1: What about https://github.com/kosmtik/kosmtik-mbtiles-export?

docs/DEPLOY_PODMAN.md
```diff
@@ -0,0 +1,117 @@
...
+dc up -d db
+dc run --rm import
+```
+
+Import tuning lives in `.env` (created on first import). For a big extract raise
+`OSM2PGSQL_CACHE` / `OSM2PGSQL_NUMPROC` and `PG_MAINTENANCE_WORK_MEM` to match the
+box, then re-run `dc run --rm import` (it re-imports from scratch — see DOCKER.md).
+The DB persists in the `db-data` named volume; re-import only on data updates.
+
+## 4. Render tiles to MBTiles
```
Phyks, line 68: The doc could use a proofreading, this seems to belong to the STATIC_TILES.md file rather?

Phyks (commented):
Hey! Thanks for the PR! Left a couple coments, happy to discuss and get this merge.

docker-compose.podman.yml
thof: Yes, I fully tested rootless podman setup on linux box and docker on macOS.

Dockerfile.db
thof: The official image only supports AMD64 at the moment. The one I used comes from the core developer of postgis though.

scripts/render-tiles.js
thof: I didn't know about it, also I haven't had very good experience with kosmtik. This rendered uses mapnik.

Dockerfile
thof: mapnik-core is available but it doesn't include mapnik-config. That's a bit hacky way to pass preinstall without touching the other repo.

docs/DEPLOY_PODMAN.md
thof: Not sure about this. It basically describes a full pipeline, i.e. what to run to have a working tile server with docker/podman only. So it's more like a recipe from a cookbook that you can use to get a very specific result. We can move it to a specific dir like `demo` or something like that.

Issue #722, opened 2026-07-06
Support for the surface=laterite tag


ftrebien:
[surface=laterite](https://wiki.openstreetmap.org/wiki/Tag:surface=laterite) is now an approved OSM surface tag. It describes unpaved roads surfaced with lateritic soil that lose traction in wet conditions. Would you be willing to add support for it? It seems like it fits alongside the unpaved surfaces already classified as `surface_type=mtb` in `views.sql`.

Phyks:
Hey! Sure, happy to take a PR for this. :)

PR #723, opened 2026-07-18
Render surface=laterite as unpaved

ftrebien:
Addresses #722: classifies `surface=laterite` as existing unpaved surfaces (`surface_type=mtb`) in `views.sql`.


Phyks:
Thanks!

